PR #2, opened 2020-02-27
Feature/base

titouanmathis:
This PR sets the base structure of the project: 

- Add a `Base` abstract class (see the [proposal](https://gist.github.com/titouanmathis/c0fa9b02f1e98854c459466a2b16a752) for detailed informations)
- Add an `EventManager` abstract class
- Add a `hasMethod(object, methodName)` utils
- Add services of type `use...` for common stuff : scroll, resize, pointer, requestAnimationFrame
- Add a [demo](https://deploy-preview-2--studiometa-js-toolkit-demo.netlify.com/) 

**To do**

- [ ] Add documentation for the new stuff


src/services/pointer.js
```diff
@@ -0,0 +1,189 @@
...
+    });
+    document.addEventListener('touchstart', this.downHandler, {
+      passive: true,
+    });
+    document.addEventListener('mouseup', this.upHandler, {
+      passive: true,
+    });
+    document.addEventListener('touchend', this.upHandler, {
+      passive: true,
+    });
```
perruche: Should we set a eslint-disable here to have all our listeners written the same way ?

src/services/pointer.js
```diff
@@ -0,0 +1,189 @@
...
+   * @return {void}
+   */
+  updateValues(event) {
+    this.yLast = this.y;
+    this.xLast = this.x;
+
+    // Check scroll Y
+    if (((event.touches || [])[0] || event || {}).clientY !== this.y) {
+      this.y = ((event.touches || [])[0] || event || {}).clientY;
+    }
```
perruche: This is not an issue since we propably won't change those parts, but i find this is quite difficult to read/understand

src/services/pointer.js
```diff
@@ -0,0 +1,189 @@
...
+      },
+      delta: {
+        x: this.x - this.xLast,
+        y: this.y - this.yLast,
+      },
+      progress: {
+        x: this.x / window.innerWidth,
+        y: this.y / window.innerHeight,
+      },
+      max: {
```
perruche: Smart to have that window size available. 
Should we go further and expose those also in a more global scope ?

src/services/pointer.js
```diff
@@ -0,0 +1,189 @@
+/* eslint-disable require-jsdoc */
+import Service from '../abstracts/Service';
+import { throttle, debounce } from '../utils';
+import useRaf from './raf';
+
+/**
+ * Pointer service
+ *
+ * ```
+ * import scroll from '@studiometa/js/services/pointer';
```
perruche: import pointer 😄

src/services/raf.js
```diff
@@ -0,0 +1,67 @@
...
+ * ```
+ * import tick from '@studiometa/js/services/tick';
+ * tick.add(id, handler);
+ * tick.remove(id);
+ * tick.props;
+ * ```
+ */
+class Raf extends Service {
+  /** @type {Boolean} Whether the loop is running or not. */
+  isLooping = false;
```
perruche: maybe `isTicking` is more appropirate ?

demo/index.html
```diff
@@ -0,0 +1,245 @@
...
+      onload="this.setAttribute('media', 'all');this.removeAttribute('onload')"
+    />
+  </head>
+
+  <body>
+    <main>
+      <!-- BEGIN Accordion component-->
+      <div class="accordion" data-component="Accordion">
+        <div data-component="AccordionItem" tabindex="0" class="accordion-item">
+          <p data-ref="AccordionItem.title" class="accordion-item__title">
```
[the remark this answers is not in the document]
titouanmathis: We need the component's name in the `data-ref` attribute to make sure the ref is tied to the correct component. The ref name [is extracted](https://github.com/studiometa/js-toolkit/blob/feature%2Fbase/src/abstracts/Base.js#L32) with a `string.replace(name + '.', '')` : 

```js
const refName = $ref.dataset.ref.replace(`${name}.`, '');
```

.eslintrc.js
```diff
@@ -1,3 +1,25 @@
...
+  extends: '@studiometa/eslint-config',
+  globals: {
+    window: false,
+    document: false,
+    requestAnimationFrame: false,
+    IntersectionObserver: false,
+    Image: false,
+  },
+  rules: {
+    'class-methods-use-this': 'off',
```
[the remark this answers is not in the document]
titouanmathis: Yes, we could update our config according to our usage. Do you think you could open a PR on the subject?

src/abstracts/Base.js
```diff
@@ -0,0 +1,284 @@
...
+  /**
+   * Class constructor where all the magic takes place
+   * @param  {Object}    options An option object
+   * @return {Base}         The created instance
+   */
+  constructor(element) {
+    super();
+
+    if (!this.config) {
+      throw new Error('The `config` getter must be defined.');
```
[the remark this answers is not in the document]
titouanmathis: A link to the documentation maybe?

src/abstracts/Base.js
```diff
@@ -0,0 +1,284 @@
...
+    if (!this.config.name) {
+      throw new Error('The `config.name` property is required.');
+    }
+
+    this.$isMounted = false;
+    this.$id = `${this.config.name}-${nanoid()}`;
+    this.$el = element || document.querySelector(this.config.el);
+
+    if (!this.$el) {
+      throw new Error('Unable to find the root element.');
```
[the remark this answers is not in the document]
titouanmathis: The root element can have multiple children, but you can't instantiate a component with an array of elements, so `this.$el` will always be one element, either passed as a parameter in the constructor, or via the `document.querySelector(this.config.el)` method.

src/abstracts/Service.js
```diff
@@ -0,0 +1,185 @@
...
+    test: descriptor => typeof descriptor.value === 'function',
+    error: 'The `kill` method must be implemented.',
+  },
+  /**
+   * The `props` property must be a getter
+   * @type {Object}
+   */
+  props: {
+    test: descriptor => typeof descriptor.get === 'function',
+    error: 'Foo',
```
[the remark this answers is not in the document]
titouanmathis: Indeed 🙂

src/services/pointer.js
titouanmathis: We can get the props of any service by importing it: 

```js
import { usePointer } from '@studiometa/js-toolkit/services';

const { props } = usePointer();
console.log(props().max);
```

src/services/pointer.js
titouanmathis: It is an artifact of our 80 characters per line rule, this is fixed 😉

src/services/pointer.js
titouanmathis: I don't really understand it anymore ^^

Creating variable comes with a cost, so I suggest to leave it untouched for the moment.

ptt-homme: Ok to leave it as it for now, but maybe add a comment with the old JavaScript form + variables to illustrate the code below;-)

src/abstracts/Base.js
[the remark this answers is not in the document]
perruche: Yes that is a good starting point !

.eslintrc.js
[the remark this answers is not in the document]
perruche: the PR is open !

src/abstracts/Base.js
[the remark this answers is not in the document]
titouanmathis: The doc will come in a future PR, I will update this error at the same time.

PR #10, opened 2020-04-14
feature/SM-459-accordion

w-jerome:
Adds the Accordion component.

It allows you to create an opening and closing animation on successive HTML elements.

It has some features :
- Accessibility management
- To have items open by default
- To close items when you open one
- To choose the animation time


docs/components/Accordion.md
```diff
@@ -0,0 +1,641 @@
...
+
+By default, the `AccordionItem` component offers an animation. Animations are managed individually for each item.
+
+If you want to disable an animation you will have to switch `animation: false` to the item. And if you want to set your animation duration, you'll have to pass the `animationDuration: 500` parameter (the time is in milliseconds).
+
+#### Disable animation
+
+<Preview>
+  <div class="accordion bg-white" data-component="Accordion">
+    <article class="border -mb-px" data-component="AccordionItem" data-options='{ "animation": false }'>
```
perruche, line 304: i am wondering if this options is useful ? I do not know if you will ever disable the animation while using this component

docs/components/Accordion.md
```diff
@@ -0,0 +1,641 @@
...
+        </p>
+      </div>
+    </div>
+  </article>
+</div>
+```
+
+#### Choice animation duration
+
+<Preview>
```
perruche, line 394: Could be good to add the default value of `animationDuration` in the doc here

src/components/Accordion.js
```diff
@@ -0,0 +1,369 @@
...
+
+  /**
+   * Initialize the component's behaviours.
+   *
+   * @return {AccordionItem} The current instance.
+   */
+  mounted() {
+    this.$refs.btn.setAttribute('id', this.$id);
+    this.$refs.btn.addEventListener('click', this);
+    this.$refs.content.setAttribute('aria-labelledby', this.$id);
```
perruche, line 27: 👍

docs/components/Accordion.md
```diff
@@ -0,0 +1,641 @@
...
+
+### Items auto-close
+
+To close items other than the one you just clicked, just pass the `itemAutoClose: true` parameter. This option is passed to the `Accordion` component.
+
+<Preview>
+  <div class="accordion bg-white" data-component="Accordion" data-options='{ "itemAutoClose": true }'>
+    <article class="border -mb-px" data-component="AccordionItem">
+      <header>
+        <button type="button" class="w-full p-4 text-left cursor-pointer" data-ref="AccordionItem.btn">
```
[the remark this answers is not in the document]
w-jerome: That's right. In this case on this component, this HTML element does not only `open`, it also `close`. So I suggest we call it `toggle` ?

docs/components/Accordion.md
w-jerome: This option was quick and easy to implement.

But let's imagine that the opening or closing of the accordion is "managed" by an external element.

Maybe this option can be useful in this case because it doesn't `repaint` at 60 fps something outside the viewport and therefore a drop in performance.

docs/components/Accordion.md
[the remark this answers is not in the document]
perruche: What is your take on this @titouanmathis  ? 
It's important that we have consistence and logic in the naming convention

docs/components/Accordion.md
perruche: yeah there is maybe no point in removing it, it won't be use a lot, but that's what default value are here :)

src/components/Accordion.js
```diff
@@ -0,0 +1,369 @@
...
+  /**
+   * Initialize the component's behaviours.
+   *
+   * @return {Accordion} The current instance.
+   */
+  mounted() {
+    if (this.$options.autoClose) {
+      this.unbindOpen = this.$children.AccordionItem.map((item, index) => {
+        return item.$on('open', () => {
+          this.openHandler(index);
```
titouanmathis, line 336: I think the `openHandler` method can be squizzed and its behavior directly here.

src/components/Accordion.js
```diff
@@ -0,0 +1,369 @@
...
+    return this;
+  }
+
+  /**
+   * Accordion open event.
+   *
+   * @param {Integer} index The AccordionItem index.
+   */
+  openHandler(index = 0) {
+    this.$children.AccordionItem.filter((el, i) => index !== i).forEach(item => {
```
titouanmathis, line 363: We can filter both the index and the state of the item:

```js
this.$children.AccordionItem
  .filter((item, i) => index !== i && item.isOpen)
  .forEach(item => item.close());
```

src/components/Accordion.js
```diff
@@ -0,0 +1,369 @@
+/* eslint-disable max-classes-per-file */
+import { Base } from '../abstracts';
```
titouanmathis, line 2: Prefer importing from the original file to improve the tree-shaking capabilities of the toolkit:

```js
import Base from '../abstracts/Base';
```

src/components/Accordion.js
```diff
@@ -0,0 +1,369 @@
...
+      this.$refs.content.style.display = 'none';
+    }
+
+    this.$emit('afterAnimation', this);
+  }
+
+  /**
+   * AccordionItem register event.
+   */
+  handleEvent(event) {
```
titouanmathis, line 163: I think we can drop this event dispatcher method as every method is bound to the instance. You will be able to bind the `clickHandler` method directly:

```js
this.$el.addEventListener('click', this.clickHandler);
// ...
this.$el.removeEventListener('click', this.clickHandler);
```

titouanmathis:
A version of the Accordion component with transition has been added in #23, I am closing this.

PR #11, opened 2020-05-05
Feature/modal component

titouanmathis:
This PR adds a new Modal component along with the following utilities:

- `keyCodes` to easily find which key has been pressed
- `nextFrame` to delay the execution of a function to the next frame
- `tabTrap` to generate `trap` and `untrap` method to be used to trap the tab navigation inside a given element

The following modifications were made to the `Base` class:

- `data-ref` on a child component will resolve to the child component's instance
- a component's instance is attached to a `__base__` property on its root element
- all methods added to a component are autobinded to its instance (no more `this.foo = this.foo.bind(this)`)

See the [updated documentation](https://deploy-preview-11--studiometa-js-toolkit.netlify.app/components/Modal.html) for more explanation and examples.


src/components/Modal.js
```diff
@@ -0,0 +1,185 @@
...
+   * Initialize the component's behaviours.
+   *
+   * @return {Modal} The current instance.
+   */
+  mounted() {
+    this.isOpen = false;
+    this.open = this.open.bind(this);
+    this.close = this.close.bind(this);
+    this.keydownHandler = this.keydownHandler.bind(this);
+    this.$refs.btn.addEventListener('click', this.open);
```
perruche, line 74: ref.btn = event open 
ref.close = event close 
Should this be the same names ? Do we already have the same naming convention for other components ?

src/components/Modal.js
```diff
@@ -0,0 +1,185 @@
...
+   *
+   * @return {Modal} The Modal instance.
+   */
+  open() {
+    this.$refs.modal.setAttribute('aria-hidden', 'false');
+    document.documentElement.style.overflow = 'hidden';
+    document.addEventListener('keydown', this.keydownHandler);
+
+    // Add "open" classes to refs
+    Object.entries(this.$options.openClass).forEach(([ref, classes]) => {
```
perruche, line 111: is it performant to run 4 loops here ? 
i know this is not suppose to add a million classes but could be good to check ? 

Also is it necessary to have option to add inline styles ?  
if we can add a class, we can apply style to it ? To simplify the usage of the component ?

src/utils/keyCodes.js
```diff
@@ -0,0 +1,10 @@
+export default {
+  ENTER: 13,
+  SPACE: 32,
+  TAB: 9,
+  ESC: 27,
+  LEFT: 37,
+  UP: 38,
+  RIGHT: 39,
+  DOWN: 40,
+};
```
perruche, line 10: 👍

src/utils/tabTrap.js
```diff
@@ -0,0 +1,75 @@
...
+  '[contenteditable]:not([tabindex^="-"]):not([inert])',
+  '[tabindex]:not([tabindex^="-"]):not([inert])',
+];
+
+/**
+ * Use a trap/untrap tabs logic.
+ *
+ * @return {Object} An object containing the trap and untrap methods.
+ */
+export default function useTabTrap() {
```
perruche, line 22: This concept could use some more documentation i think, i didn't quite understand the concept without reading the code

src/components/Modal.js
titouanmathis: I did some tests with Tailwind for the Tabs component, and having the possibility to set both classes and styles can be helpful at times.

As for performance, a `console.time` + `console.timeEnd` test gave me between 0 and 6ms of execution time. I think we can keep it that way for now and adjust this behaviour later if it is indeed too greedy.

titouanmathis:
> Is it possible to think of a programatic way to open the modal ?
> Use case like open the Modal after Xsec on a page

In order to keep things simple, a programmatic usage of the modal should be made in a parent component by using the `open` and `close` methods exposed by the component. 

```js
import { Base, Modal } from '@studiometa/js-toolkit';

class MyPage extends Base {
  get config() {
    return {
      name: 'MyPage',
      components: { Modal },
    };
  }

  mounted() {
    setTimeout(() => this.$children.Modal[0].open(), 500);
  }
}
```

perruche:
> > Is it possible to think of a programatic way to open the modal ?
> > Use case like open the Modal after Xsec on a page
> 
> In order to keep things simple, a programmatic usage of the modal should be made in a parent component by using the `open` and `close` methods exposed by the component.
> 
> ```js
> import { Base, Modal } from '@studiometa/js-toolkit';
> 
> class MyPage extends Base {
>   get config() {
>     return {
>       name: 'MyPage',
>       components: { Modal },
>     };
>   }
> 
>   mounted() {
>     setTimeout(() => this.$children.Modal[0].open(), 500);
>   }
> }
> ```

👍 , is this exemple is worth adding to the component documentation ?

src/components/Modal.js
titouanmathis: Fixed!  🙂 

We now have `ref.open` and `ref.close` which can both have multiple occurences if needed.

src/utils/tabTrap.js
titouanmathis: I added a link to an article explaining its usage in the documentation: https://deploy-preview-11--studiometa-js-toolkit.netlify.app/utils/#tabtrap

titouanmathis:
> is this exemple is worth adding to the component documentation ?

I added the example to the doc and improved the resolution of refs. If a `data-ref="Parent.foo"` is added to a component's root element, `this.$refs.foo` will return the component's instance.

```html
<div data-component="Parent">
  <div data-component="Child" data-ref="Parent.child">
    ...
  </div>
</div>
```

```js

class Child extends Base {
  get config() {
    return {
      name: 'Child',
    };
  }

  foo() {
    console.log('foo');
  }
}

class Parent extends Base {
  get config() {
    return {
      name: 'Parent',
      components: { Child },
    };
  }

  mounted() {
    console.log(this.$refs.child.$options.name); // 'Child'
    this.$refs.child.foo(); // 'foo'
  }
}

File: src/abstracts/Base.js [hunk 5]
```diff
@@ -1,4 +1,5 @@
 import nanoid from 'nanoid';
+import autoBind from 'auto-bind';
```
perruche, line 2: 🔥

titouanmathis: 🙂

demo/index.html
```diff
@@ -13,6 +13,145 @@
...
+                Lorem ipsum dolor sit amet, consectetur adipisicing elit. Praesentium sapiente exercitationem, numquam fuga molestiae autem repudiandae tempora quidem eveniet laborum beatae dolorum dicta quisquam unde consequatur natus corrupti illo? Voluptate?
+                Lorem ipsum dolor sit amet, consectetur adipisicing elit. Praesentium sapiente exercitationem, numquam fuga molestiae autem repudiandae tempora quidem eveniet laborum beatae dolorum dicta quisquam unde consequatur natus corrupti illo? Voluptate?
+                Lorem ipsum dolor sit amet, consectetur adipisicing elit. Praesentium sapiente exercitationem, numquam fuga molestiae autem repudiandae tempora quidem eveniet laborum beatae dolorum dicta quisquam unde consequatur natus corrupti illo? Voluptate?
+                Lorem ipsum dolor sit amet, consectetur adipisicing elit. Praesentium sapiente exercitationem, numquam fuga molestiae autem repudiandae tempora quidem eveniet laborum beatae dolorum dicta quisquam unde consequatur natus corrupti illo? Voluptate?
+                <a href="#">Link</a>
```
perruche, line 145: Is it worth adding something like data-ref="Modal.initialFocus"` to choose the element of the modal we want to focus when it is open ? (for some we might want close button, for other the text first text input)

titouanmathis: Added in 61e4e84 via the `autofocus` attribute which can be added on any child of the `[data-ref="Modal.modal"]` element.

The documentation is up to date with this addition: https://deploy-preview-11--studiometa-js-toolkit.netlify.app/components/Modal.html#autofocus

docs/components/Modal.md
```diff
@@ -0,0 +1,618 @@
...
+This ref is placed inside the container and is used to close the modal.
+
+#### `Modal.content`
+
+This ref will hold the modal's dynamic content.
+
+### Options
+
+::: tip
+Options can be defined per component via the `data-options` attribute or by extending the Tabs class.
```
perruche, line 575: i guess it should be Modal class ?

docs/components/Modal.md
```diff
@@ -0,0 +1,618 @@
...
+An object which keys are refs names and values are classes (`String`) that will be added to the refs when the modal opens.
+
+#### `openStyle`
+
+- Type: `Object`
+- Default: `{}`
+
+An object which keys are refs names and values are style definition (`CSSStyleDeclaration`) that will be added to the refs when the modal opens.
+
+#### `closedClass`
```
perruche, line 606: I'm wondering if this should be `closeStyle` & `closeClass` without the D as in `openClass` & `openStyle`, what's you opinion on this @titouanmathis ?

src/components/Modal.js
```diff
@@ -0,0 +1,199 @@
...
+
+    const open = Array.isArray(this.$refs.open) ? this.$refs.open : [this.$refs.open];
+    open.forEach(btn => btn.addEventListener('click', this.open));
+
+    const close = Array.isArray(this.$refs.close) ? this.$refs.close : [this.$refs.close];
+    close.forEach(btn => btn.addEventListener('click', this.close));
+
+    this.$refs.overlay.addEventListener('click', this.close);
+
+    this.close();
```
perruche, line 81: It is a bit weird to run `Modal.close()` on mounted no ?

src/components/Modal.js
```diff
@@ -0,0 +1,199 @@
...
+
+    if (this.$options.move) {
+      const target = document.querySelector(this.$options.move) || document.body;
+      target.appendChild(this.$refs.modal);
```
perruche, line 85: it's also weird to move the modal on mounted ? The main point i see about moving the modal is for screen readers: you do not want your modal on top of your html markup, but you want it on top when it's open so the screen reader read the correct content. (But maybe this whole point is handle by the tabTrap ?)

I guess i find it stange to move your modal as soon as the code run ? why not put the Modal at the desire place directly ? Is this for seo concerns ?

src/components/Modal.js
```diff
@@ -0,0 +1,199 @@
...
+      setStyles(this.$refs[ref], styles, 'remove');
+    });
+
+    if (this.$options.autofocus && this.$refs.modal.querySelector(this.$options.autofocus)) {
+      saveActiveElement();
+      this.$refs.modal.querySelector(this.$options.autofocus).focus();
+    }
+
+    this.isOpen = true;
+    this.$emit('open');
```
perruche, line 145: The whole `$emit` part is not cover in the doc ATM i think ? This is an event we can listen to right ?

docs/components/Modal.md
titouanmathis: It is gramatically more correct to have the `d`:

- close style → fermer le style
- closed style → style fermé
- open style → style ouvert

😉

docs/components/Modal.md
titouanmathis: Fixed, thanks! 👍

src/components/Modal.js
titouanmathis: Yes, you can listen to the events on the component instance:

```js
const modal = new Modal(document.querySelector('[data-component="Modal"]');

modal.on('open', () => {
  // Do stuff when the modal opens.
});

modal.on('close', () => {
  // Do stuff when the modal closes.
});
```

I added the events to the documentation.

src/components/Modal.js
titouanmathis: It was added to make sure the style and class of the closed state were added to the elements. 

It might be removed, but we will have to add closed styles and classes in the markup.

titouanmathis: I think we can keep it for now and adjust this behavior later on.

perruche: Does this mean the modal is visible until the script run and hide the modal ?

src/components/Modal.js
titouanmathis: > why not put the Modal at the desire place directly ?

The component is made of a button (or link) and the modal grouped inside a single HTML parent element, so: 

- we can not separate the two of them by placing the modal markup somewhere else in our page
- but we can manage our modal template in a single Twig (or other templating language) file
and include/embed the file where it is needed

Maybe we do not want the modal HTML element to stay next to its button (CSS conflict, SEO, etc.), so I added a `move` options which would allow us to chose where the modal HTML element should be placed. It works a little bit like [Vue's portal component](https://portal-vue.linusb.org/).

src/components/Modal.js
titouanmathis: It is a safeguard which will add the closed state styles and classes if you forgot to add them to the component's markup. You might have a flash of unstyled modal if you did forget.

PR #12, opened 2020-05-11
feature/component-instance

w-jerome:
## Problem

When using the `data-ref="my-component"` attribute on a component, you cannot access the API of the component. It only returns the DOM element. So we can't interact with the API of the component.

### Example 

```html
<!-- my-page.html -->

<div data-component="MyPage">
    <button data-ref="MyPage.externalModalOpenBtn">
        Open modal
    </button>

    <div data-component="Modal" data-ref="MyPage.modal">
        ...
    </div>
</div>
```

```js
// my-page.js

class MyPage extends Base {
    get config() {
        return {
            name: 'MyPage',
            components: {
                Modal,
            },
        };
    }

    mounted() {
        this.$refs.externalModalOpenBtn.addEventListener('click', () => {
            this.$refs.modal.open(); // ERROR undefined
        });

        return this;
    }
}
```

## Solution

Add the instance of the component in the properties of the parent element.

### Example

```js
// my-page.js

class MyPage extends Base {
    get config() {
        return {
            name: 'MyPage',
            components: {
                Modal,
            },
        };
    }

    mounted() {
        this.$refs.externalModalOpenBtn.addEventListener('click', () => {
            this.$refs.modal.$instance.open(); // 🎉
        });

        return this;
    }
}
```


perruche:
I think this enhancement was also done by @titouanmathis on #11 

[Commit ref](https://github.com/studiometa/js-toolkit/pull/11/commits/d567cdcfae381975bd2671b971fb13beaad33097)

PR #13, opened 2020-05-18
Feature/key service

titouanmathis:
This PR adds the following:

- A `useKey` service to easily manage keyboard events (see [doc.](https://deploy-preview-13--studiometa-js-toolkit.netlify.app/services/key.html))
- Documentation for all the available services (see [doc.](https://deploy-preview-13--studiometa-js-toolkit.netlify.app/services/pointer.html))


docs/services/scroll.md
```diff
@@ -0,0 +1,83 @@
...
+- Type: `Number`
+
+The current vertical scroll position.
+
+### `changed`
+
+- Type: `Object`
+- Interface: `{ x: boolean, y: boolean }`
+
+For both axis, whether or not the current position has changed since the last one.
```
perruche, line 55: What is the difference between `changed` and `last` ? Is `changed` the current state value ?

src/services/key.js
```diff
@@ -0,0 +1,102 @@
+import Service from '../abstracts/Service';
+import { keyCodes } from '../utils';
+
+/**
+ * Scroll service
+ *
+ * ```
+ * import { useScroll } from '@studiometa/js-toolkit/services';
+ * const { add, remove, props } = useScroll();
```
perruche, line 9: The comments on the file are refering to the scroll service

docs/services/scroll.md
titouanmathis: `changed` is a boolean, it can be used to check if the `x` or `y` scroll has changed since the last callback. `last` is the previous value, it is used to define the `delta` values: 

```
delta.x = x - last.x;
delta.y = y - last.y;
```

PR #21, opened 2020-06-18
Feature/async components

titouanmathis:
This PR adds support for async components:

```js

class Page extends Base {
  get config() {
    return {
      name: 'Page',
      components: {
        Component: () => import('./components/Component.js'),
      },
    };
  }
}
```

The async component will only be loaded if one or more `[data-component="Component"]` elements are present in the page.

--- 

**Changelog entry:**

### Added
- Add support for async component (#21)

### Remove
- Remove the `el` property in a component's configuration, a component can not anymore be instantiated without a root element as parameter (#21)


src/abstracts/Base.js
```diff
@@ -64,17 +64,33 @@ function getChildren(element, components) {
   }
 
   return Object.entries(components).reduce((acc, [name, ComponentClass]) => {
-    const { config } = ComponentClass.prototype || {};
-    const selector = config.el || `[data-component="${name}"]`;
+    const selector = `[data-component="${name}"]`;
```
perruche, line 67: I don't mind this change since it reduce the complexity of creating a component. However i think this is a breaking change. 
Should we add a console warning if `config.el` is set ? Could be usefull if we update the toolkit on a project

titouanmathis: The package is still in the alpha phase, so breaking change can be introduced, but a warning might be a good idea. I'll add one.

PR #25, opened 2020-08-06
Feature/custom component selector

titouanmathis:
This PR adds support for custom selector when defining children components. Components can now be defined with the following methods:

```js
import ComponentOne from './ComponentOne';
import ComponentThree from './ComponentThree';

class App extends Base {
  get config() {
    return {
      name: 'App',
      components: {
        // Direct reference, ComponentOne instances will be mounted 
        // on every `[data-component="ComponentOne"]` elements.
        ComponentOne,
        // Async loading, the ComponentTwo component will be loaded only if there
        // is one or more `[data-component="ComponentTwo"]` element in the DOM.
        ComponentTwo: () => import('./ComponentTwo'),
        // Custom selector, ComponentThree instances will be mounted 
        // on every `.custom-selector` elements.
        '.custom-selector': ComponentThree,
        // Custom selector with async loading, the ComponentFour component will be loaded 
        // only if there is one or more `.other-custom-selector` element in the DOM.
        '.other-custom-selector': () => import('./ComponentFour'),
      },
    };
  }
}
```

---

## Changed
- Update documentation with demo link and introduction (7d1f908)
- Update the API documentation with the custom selector support (6415ce4)

## Added
- Add an example of the custom selector in the demo (dde316f)
- Add support for custom selector in component definition (f36d68a)


src/abstracts/Base.js
```diff
@@ -76,7 +76,12 @@ function getRefs(instance, element) {
 function getChildren(instance, element, components) {
   const children = Object.entries(components).reduce((acc, [name, ComponentClass]) => {
     const selector = `[data-component="${name}"]`;
-    const elements = Array.from(element.querySelectorAll(selector));
+    let elements = Array.from(element.querySelectorAll(selector));
+
+    // If no child component found with the default selector, the name must be a DOM selector
+    if (elements.length === 0) {
+      elements = Array.from(element.querySelectorAll(name));
```
perruche, line 83: Should maybe throw error if no results for qsa(name) ? To improve the dev experience

docs/api/readme.md
```diff
@@ -8,14 +8,70 @@ next: /components/
...
+- Default: `false`
+
+Enable the `this.$log(...args)` method when `true`.
+
+#### `debug`
+
+- Type: `Boolean`
+- Default: `false`
+
+When `true`, the lifecycle hooks and services hooks will be logged to the console.
```
perruche, line 72: Great to have some documentation update in the PR 👍

demo/src/js/components/Lazyload.js
```diff
@@ -0,0 +1,56 @@
...
+        from: unloaded,
+        active,
+        to: loaded,
+      });
+      setClassesOrStyles(this.$el, loaded);
+      this.$destroy();
+    };
+    img.src = src;
+  }
+}
```
perruche, line 56: Should this be a default component available to import ?

src/abstracts/Base.js
titouanmathis: Actually, this is a feature 🙂 
It fails silently on purpose to load async components only if elements are found in the DOM.

demo/src/js/components/Lazyload.js
titouanmathis: I think it still needs some work to handle all cases : `video`, `iframe` and `picture` elements along with the `srcset` and `sizes` attributes.

perruche: Yes that's right. That's a dope component to have in the toolkit ! Can't wait 🦖

PR #27, opened 2020-08-18
Add first draft of the documentation for the breakpoint abstract classes

titouanmathis:
This PR introduces mechanisms to easily work with breakpoints when creating components:

- The `BreakpointObserver` abstract class allows a component to self mount/destroy by breakpoint
- The `BreakpointManager` abstract class allows a component to switch between different component classes by breakpoint

### DIFF
--- a/docs/abstracts/BreakpointManager.md
+++ b/docs/abstracts/BreakpointManager.md
@@ -1,0 +1,40 @@
+---
+sidebar: auto
+sidebarDepth: 5
+prev: /abstracts/
+next: /abstracts/BreakpointObserver.html
+---
+
+# BreakpointManager
+
+A component based on this class will have the capacity to switch components between different breakpoints.
+
+## Examples
+
+### Switch between component classes
+
+In the following example, the `MenuMobile` class will be mounted along the `Menu` class on small devices and destroyed on large devices. The `MenuDesktop` class will be mounted on large devices and destroyed on small ones.
+
+The root element `this.$el` of each classes will be the same.
+
+```js{9-12}
+import { BreakpointManager } from '@studiometa/js-toolkit/abstracts';
+import MenuMobile from './MenuMobile';
+import MenuDesktop from './MenuDesktop';
+
+export default class Menu extends BreakpointManager {
+  get config() {
+    return {
+      name: 'Menu',
+      breakpoints: [
+        ['xxs xs s', MenuMobile],
+        ['m l xl xxl', MenuDesktop],
+      ],
+    };
+  }
+}
+```
+
+:::tip
+See the [`resize` service documentation on breakpoints](/services/resize.html#breakpoint) for a more comprehensive view of the potential values of the `activeBreakpoint` property.
+:::
--- a/docs/abstracts/BreakpointObserver.md
+++ b/docs/abstracts/BreakpointObserver.md
@@ -1,0 +1,50 @@
+---
+sidebar: auto
+sidebarDepth: 5
+prev: /abstracts/BreakpointManager.html
+next: /components/
+---
+
+# BreakpointObserver
+
+A component based on this class will have the capacity to be mounted or destroyed when the current breakpoint changes.
+
+## Examples
+
+### Enable for the given breakpoints
+
+In the following example, the `MobileComponent` class will self mount on devices matching any of the breakpoint defined in the `activeBreakpoint` property, and self destroy on all others:
+
+```js{7}
+import { BreakpointObserver } from '@studiometa/js-toolkit/abstracts';
+
+export default class MobileComponent extends BreakpointObserver {
+  get config() {
+    return {
+      name: 'MobileComponent',
+      activeBreakpoints: 'xxs xs s'
+    };
+  }
+}
+```
+
+### Disable for the given breakpoints
+
+The same behaviour as before can be achieved by specifying all the other breakpoints in the `inactiveBreakpoints` property instead:
+
+```js{7}
+import { BreakpointObserver } from '@studiometa/js-toolkit/abstracts';
+
+export default class MobileComponent extends BreakpointObserver {
+  get config() {
+    return {
+      name: 'MobileComponent',
+      inactiveBreakpoints: 'm l xl xxl',
+    };
+  }
+}
+```
+
+:::tip
+See the [`resize` service documentation on breakpoints](/services/resize.html#breakpoint) for a more comprehensive view of the potential values of the `activeBreakpoint` property.
+:::
--- a/docs/.vuepress/config.js
+++ b/docs/.vuepress/config.js
@@ -9,6 +9,14 @@ module.exports = {
       { text: 'Guide', link: '/guide/' },
       { text: 'API', link: '/api/' },
       {
+        text: 'Abstracts',
+        link: '/abstracts/',
+        items: [
+          { text: 'BreakpointManager', link: '/abstracts/BreakpointManager.html' },
+          { text: 'BreakpointObserver', link: '/abstracts/BreakpointObserver.html' },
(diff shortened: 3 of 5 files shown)


docs/abstracts/BreakpointManager.md
```diff
@@ -0,0 +1,40 @@
...
+In the following example, the `MenuMobile` class will be mounted along the `Menu` class on small devices and destroyed on large devices. The `MenuDesktop` class will be mounted on large devices and destroyed on small ones.
+
+The root element `this.$el` of each classes will be the same.
+
+```js{9-12}
+import { BreakpointManager } from '@studiometa/js-toolkit/abstracts';
+import MenuMobile from './MenuMobile';
+import MenuDesktop from './MenuDesktop';
+
+export default class Menu extends BreakpointManager {
```
perruche, line 25: I find it strange to not extend `Base` anymore

docs/abstracts/BreakpointManager.md
```diff
@@ -0,0 +1,40 @@
...
+  get config() {
+    return {
+      name: 'Menu',
+      breakpoints: [
+        ['xxs xs s', MenuMobile],
+        ['m l xl xxl', MenuDesktop],
```
perruche, line 31: This is real nice

docs/abstracts/BreakpointObserver.md
```diff
@@ -0,0 +1,50 @@
...
+The same behaviour as before can be achieved by specifying all the other breakpoints in the `inactiveBreakpoints` property instead:
+
+```js{7}
+import { BreakpointObserver } from '@studiometa/js-toolkit/abstracts';
+
+export default class MobileComponent extends BreakpointObserver {
+  get config() {
+    return {
+      name: 'MobileComponent',
+      inactiveBreakpoints: 'm l xl xxl',
```
perruche, line 42: This will probably need to throw a conprehensive error if a breakpoint is both in the active and inactive string

perruche (commented):
Real nice addition. I'm just wondering about the extend class, is it too much to have it as a feature in the `Base` class ?

docs/abstracts/BreakpointManager.md
lusimeon: I think the inheritance isn't the right pattern. Is it possible to use composition (something similar to [React Hooks](https://fr.reactjs.org/docs/hooks-intro.html) for example) ?
```js
import { BreakpointManager } from '@studiometa/js-toolkit/abstracts';
class Menu { … }
export default BreakpointManager(Menu);
```

docs/abstracts/BreakpointObserver.md
titouanmathis: An error is thrown when you configure both the `activeBreakpoints` and `inactiveBreakpoints` at the same time. The message says that the two options are not compatible.

docs/abstracts/BreakpointManager.md
titouanmathis: > I find it strange to not extend Base anymore

I would like to avoid having the `Base` class being a Swiss army knife as it would become a nightmare to maintain and test. I find it simpler to extend its functionality with small classes. The `BreakpointManager` and `BreakpointObserver` classes are not compatible, implementing their behavior in one single class would add too much complexity to the code base.

I actually find that different classes can help browse and understand the code quickly, for example: 

- if the component is extending `BreakpointObserver`, I can set the `activeBreakpoints` property
- if the component is extending `BreakpointManager`, I know that I will have to look for different classes based on the breakpoint I am working on

> I think the inheritance isn't the right pattern. Is it possible to use composition (something similar to React Hooks for example) ?

Interesting point, but as far as I could think about it, I don't see any way for now to not use inheritance. We can discuss this in more details if you want 🙂 

I added the two classes to the PR ([`BreakpointManager`](https://github.com/studiometa/js-toolkit/blob/feature%2Fbreakpoint-abstracts-classes/src/abstracts/BreakpointManager.js) and [`BreakpointObserver`](https://github.com/studiometa/js-toolkit/blob/feature%2Fbreakpoint-abstracts-classes/src/abstracts/BreakpointObserver.js)) and added examples to [the demo](https://deploy-preview-27--studiometa-js-toolkit-demo.netlify.app/).

titouanmathis:
I did some test and it would indeed be practical to have the `BreakpointObserver` class merged into the `Base` to be able to easily enable/disable a component based on the breakpoint without having to create a new class. 

For example, I tried to disable the `Modal` component from the toolkit on small devices. I can not do this without creating a new component based on `BreakpointObserver` to wrap the `Modal` component. 

I will try to find a nice implementation.

perruche:
> I did some test and it would indeed be practical to have the `BreakpointObserver` class merged into the `Base` to be able to easily enable/disable a component based on the breakpoint without having to create a new class.
> 
> For example, I tried to disable the `Modal` component from the toolkit on small devices. I can not do this without creating a new component based on `BreakpointObserver` to wrap the `Modal` component.
> 
> I will try to find a nice implementation.

Too bad interfaces aren't a thing in javascript cause something like
`class Foo extends Base implements BreakpointObserver {}` would be awesome

titouanmathis:
I updated the proposal and converted the 2 abstract classes into decorators. See the following links for more informations:

- Updated documentation: 
  - [`withBreakpointObserver`](https://deploy-preview-27--studiometa-js-toolkit.netlify.app/decorators/withBreakpointObserver.html)
  - [`withBreakpointManager`](https://deploy-preview-27--studiometa-js-toolkit.netlify.app/decorators/withBreakpointManager.html)
- Updated demo: 
  - [Demo preview](https://deploy-preview-27--studiometa-js-toolkit-demo.netlify.app/) 
  - Sources: [app.js](https://github.com/studiometa/js-toolkit/blob/feature/breakpoint-abstracts-classes/demo/src/js/app.js#L17), [BreakpointObserverDemo.js](https://github.com/studiometa/js-toolkit/blob/feature/breakpoint-abstracts-classes/demo/src/js/components/BreakpointObserverDemo.js) and [BreakpointManagerDemo/index.js](https://github.com/studiometa/js-toolkit/blob/feature/breakpoint-abstracts-classes/demo/src/js/components/BreakPointManagerDemo/index.js)

docs/decorators/withBreakpointObserver.md
```diff
@@ -0,0 +1,75 @@
+---
+sidebar: auto
+sidebarDepth: 5
+prev: /decorators/withBreakpointManager.html
+next: /components/
+---
+
+# withBreakpointObserver
+
+Use this decorator to create a that will have the capacity to be mounted or destroyed when the current breakpoint changes.
```
perruche, line 10: "to create a **Class**"

docs/decorators/withBreakpointObserver.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Examples
+
+### Enable for the given breakpoints
+
+In the following example, the `MobileComponent` class will self mount on devices matching any of the breakpoint defined in the `activeBreakpoint` property, and self destroy on all others:
```
perruche, line 16: Missing the s in `activeBreakpoints`

docs/decorators/withBreakpointObserver.md
```diff
@@ -0,0 +1,75 @@
...
+import { Base, Modal } from '@studiometa/js-toolkit';
+import { withBreakpointObserver } from '@studiometa/js-toolkit/decorators';
+
+class App extends Base {
+  get config() {
+    return {
+      name: 'App',
+      components: {
+        Modal,
+        ResponsiveModal: withBreakpointObserver(Modal),
```
perruche, line 64: That's a great behavior 👍

PR #29, opened 2020-09-01
Feature/decorator intersection observer

titouanmathis:
Add a decorator which facilitate the use of the `IntersectionObserver` API within a component.

The `Lazyload` and `Skew` components from the demo have been updated to use this decorator as examples.

- [Demo preview](https://deploy-preview-29--studiometa-js-toolkit-demo.netlify.app/)
- [Doc preview](https://deploy-preview-29--studiometa-js-toolkit.netlify.app/)


src/decorators/withIntersectionObserver.js
```diff
@@ -0,0 +1,50 @@
...
+     * Create an observer when the class in instantiated.
+     *
+     * @param  {HTMLElement} element The component's root element.
+     * @return {Base}                The class instace.
+     */
+    constructor(element) {
+      super(element);
+
+      if (!this.intersected || typeof this.intersected !== 'function') {
+        return this;
```
perruche, line 27: Maybe throw an warning/error message to explain that the component need to have an `intersected` method

src/decorators/withIntersectionObserver.js
```diff
@@ -0,0 +1,50 @@
...
+ * @return {Array}        An array of number.
+ */
+function createArrayOfNumber(length = 10) {
+  return [...new Array(length)].map((val, index) => index / length);
+}
+
+/**
+ * IntersectionObserver decoration.
+ */
+export default (BaseClass, defaultOptions = { threshold: createArrayOfNumber(100) }) =>
```
perruche, line 15: this is smart 👍

src/decorators/withIntersectionObserver.js
```diff
@@ -0,0 +1,50 @@
+import { debug } from '../abstracts/Base';
```
perruche, line 1: This should probably be remove in the release no ?
If set debug to true on the componant using decorators you'll end up with the same log results right ?

perruche (commented):
Great addition, similar behavior as the other decorators 👍 
Maybe need to add a page in the doc for `withIntersectionObserver`

perruche:
Is there a problem with the `package-lock.json` file ? A lot as been removed inside

titouanmathis:
> Is there a problem with the `package-lock.json` file ? A lot as been removed inside

It might has been regenerated using NPM 7 beta version which introduces a new format for the `package-lock.json` file. I will revert those changes.

src/decorators/withIntersectionObserver.js
titouanmathis: Yes, will do!

src/decorators/withIntersectionObserver.js
titouanmathis: The idea is to be able to debug the `intersected` method when the `debug` options is set to `true`, nothing will happen otherwise.

titouanmathis:
> Maybe need to add a page in the doc for withIntersectionObserver

Yes, it will be added later, I still need to test the decorator and see if its API is complete.

PR #32, opened 2020-09-03
Feature/terminate method

titouanmathis:
This PR adds a `$terminate()` method to remove an instance reference to be garbage collected. 

The implementation **should** work, but there is no test to confirm it.

```html
<div data-component="App">
  <div data-component="ChildComponent"></div>
  <div data-component="ChildComponent"></div>
</div>
```
```js
class ChildComponent extends Base {
  get config() {
    return {
      name: 'ChildComponent',
    };
  }

  onClick() {
    this.$terminate();
  }
}

class App extends Base {
  get config() {
    return {
      name: 'App',
      components: { ChildComponent },
    };
  }
}

const app = new App(document.querySelector('[data-component="App"]'));
const childElement = document.querySelector('[data-component="ChildComponent"]');

console.log(childElement.__base__ instanceof ChildElement); 
// true
console.log(app.$children.ChildComponent); 
// [ childComponentInstance, childComponentInstance ]

// triggers the `onClick` method of the ChildComponent
childElement.click(); 

console.log(childElement.__base__ instanceof ChildElement); 
// false
console.log(childElement.__base__); 
// 'terminated'
console.log(app.$children.ChildComponent); 
// [childComponentInstance]
```


src/abstracts/Base.js
```diff
@@ -507,6 +515,31 @@ export default class Base extends EventManager {
...
+   * Terminate a child instance when it is not needed anymore.
+   * @return {void}
+   */
+  $terminate() {
+    debug(this, '$terminate');
+
+    // First, destroy the component.
+    this.$destroy();
+
+    // Execture the `terminated` hook if it exists
```
perruche, line 529: "Execute" 😄

PR #33, opened 2020-09-03
Feature/css transform matrix

titouanmathis:
This PR adds a CSS utility function to format a transform matrix function.

```js
import matrix from '@studiometa/js-toolkit/utils/css/matrix';

const div = document.createElement('div');
div.style.transform = matrix({ scaleX: 1, scaleY: 2, skewX: 10, skewY: 11, translateX: 20, translateY: 21 });

console.log(div.style.transform);
// matrix(1, 10, 11, 2, 20, 21)
```

- [Updated documentation](https://deploy-preview-33--studiometa-js-toolkit.netlify.app/utils/#matrix)


src/utils/css/matrix.js
```diff
@@ -0,0 +1,17 @@
...
+ * @param  {Number} options.scaleY     The scale on the y axis.
+ * @param  {Number} options.skewX      The skew on the x axis.
+ * @param  {Number} options.skewY      The skew on the y axis.
+ * @param  {Number} options.translateX The translate on the x axis.
+ * @param  {Number} options.translateY The translate on the y axis.
+ * @return {String}                    A formatted CSS matrix transform.
+ */
+export default function matrix(t) {
+  // eslint-disable-next-line no-param-reassign
+  t = t || {};
```
perruche, line 14: Should we write more meaningful variable name like `transform` ?

src/utils/css/matrix.js
```diff
@@ -0,0 +1,17 @@
...
+  return `matrix(${t.scaleX || 1}, ${t.skewX || 0}, ${t.skewY || 0}, ${t.scaleY ||
```
perruche, line 15: Should we test that `t` is an object before returning ? 

And throw an error message that says something like : 
`matrix(param) must an object formated like so : { scaleX: value, skewX: value, skewY: value, scaleY: value, translateX.value, translateY: value }`

That might also allow to incorporate a more robust jest test ?

perruche (changes requested):
Side note the documentation link from the netlify buiid doesnt seem to work

titouanmathis:
> Side note the documentation link from the netlify buiid doesnt seem to work

Indeed...
It should work now 🙂

src/utils/css/matrix.js
titouanmathis: This method might seem a little "dirty" without tests and errors, but it is so by design to keep it performant as it might be used in `requestAnimationFrame` loops.

See [https://jsben.ch/fRslP](https://jsben.ch/fRslP) for the large difference between the one with test and the one without.

titouanmathis: Passing a value that is not an object will fail silently and return the default values, I will update the Jest test with this case.

src/utils/css/matrix.js
titouanmathis: Fixed!

src/utils/css/matrix.js
perruche: The perf difference is indeed huge ! 
I guess a good documentation on how to use the [`matrix()`](http://gph.is/2lEdjWA), will prevent most errors

perruche (approved):

PR #35, opened 2020-09-27
Feature/update method

titouanmathis:
This PR adds a new `$update()` instance method which can be used when inserting new HTML content or replacing some of the existing one inside a component.

```js
class App extends Base {
  /**
   * Fetching new content when click on a `data-ref="nextPage"` ref.
   */
  async onNextPageClick() {
    const content = await fetch(/* ... */);
    this.$refs.content.innerHTML = content;
    this.$update();
  }
}
```

The `$update` method will: 

- Unbind all methods from the previous content
- Mount new components (old ones will be garbage collected)
- Bind all methods to the new content (refs and children event handlers)


perruche (approved):
👍

PR #39, opened 2020-10-19
Feature/object get value deep

titouanmathis:
This PR adds a `getValueDeep` function to easily get a deep value from an object by its dotted path:

```js
import getValueDeep from '@studiometa/js-toolkit/utils/object/getValueDeep';

const obj = { foo: { bar: 'baz', array: ['one', 'two'] } };

console.log(getValueDeep(obj, 'foo.bar')); // 'baz'
console.log(getValueDeep(obj, 'foo.array')); // ['one', 'two']
console.log(getValueDeep(obj, 'foo.array.0')); // 'one'
```


lusimeon (commented):
Hey @titouanmathis, maybe we should use [optionnal chaining](https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Op%C3%A9rateurs/Optional_chaining), a [babel plugin](https://babeljs.io/docs/en/babel-plugin-proposal-optional-chaining) exists?

titouanmathis:
> Hey @titouanmathis, maybe we should use [optionnal chaining](https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Op%C3%A9rateurs/Optional_chaining), a [babel plugin](https://babeljs.io/docs/en/babel-plugin-proposal-optional-chaining) exists?

Well, that's indeed more practical and future-proof than this proposal, that did not cross my mind. Thanks for the tip @lusimeon 👍 

I am closing this PR for now, as it is already obsolete (😭).

PR #86, opened 2021-04-16
Bugfix/reset accordion on destroy

titouanmathis:
Remove styles from the AccordionItem component when it is destroyed.

Fixes #42.

PR #88, opened 2021-04-16
Release/1.1.0

### LINKED ISSUE #42: Accordion component should fully reset when destroyed
### Description
The [Accordion](https://js-toolkit.meta.fr/components/Accordion.html) component does not fully reset when destroyed. The styles added by the component are not removed. 

There is a `style="height: 0px;"` on the `data-ref="content"` which can be breaking a layout.

### How to reproduce
Init a responsive Accordion component using `withBreakpointDecorator`

### How to fix
Remove the styles when the component is `destroyed`

### LINKED ISSUE #64: Trigger the updated() hook on child componants when using $update()
Currently if you run `App.$update()` it will only trigger the `updated()` hook in the App class. Not on the child components.
It would be usefull to also trigger the `updated()` hook on child components.

This would help in case like reseting a state / rerunning an animation in a component when App is updated.

Potential solution in [src/abstracs/Base/index.js](https://github.com/studiometa/js-toolkit/blob/master/src/abstracts/Base/index.js#L172-L176) 
replacing `mountComponents(this)` with `mountOrUpdateComponents(this)`

```
 this.$on('updated', () => {
      unbindMethods.forEach((method) => method());
      unbindMethods = [...bindServices(this), ...bindEvents(this)];
      mountComponents(this);
});
```

titouanmathis:
## Added
- Add a Cursor component (#52)
- Add a `$services` property to Base instances (031d846)
- Add a `withMountWhenInView` decorator (#80)
- Add `clamp(value, min, max)` and `clamp01(value)` math utility functions (#89)

## Changed
- Improve performance of the `raf` service (and its `ticked` hook) (eed8897)

## Fixed
- Trigger children `updated` hook when a component is updated (fix #64, #77)
-  Fix nullish search params handling in the history utils (#87)
- Reset the `AccordionItem` styles when it is destroyed (fix #42, #86)


docs/api/readme.md
```diff
@@ -490,6 +490,41 @@ interface ChildrenInterface {
...
+      this.$services.disable('ticked');
+    }
+  }
+
+  ticked() {
+    // Do something on each frame...
+  }
+}
+```
+
```
perruche, line 527: praise: The $services addition is really nice, and the documentation is 💯

KeziahMoselle (approved):

docs/decorators/withMountWhenInView.md
```diff
@@ -0,0 +1,75 @@
...
+  data-option-intersection-observer='{
+    "threshold": 0.5
+  }'>
+  ...
+</div>
+```
+
+## API
+
+This decorator does not expose a specific API.
```
perruche, line 75: praise: doc is also very well put here 👍

perruche (approved):
Great work as usual 💯

PR #123, opened 2021-07-23
Feature/draggable

titouanmathis:
Add a `Draggable` component with inertia capabilities. It can be used to move an element or as a child component of another component to listen to its events.


perruche:
Thats a great addition !

titouanmathis:
The basis of the drag logic and the root element displacement are now in 2 separate components: 

- `Draggable` for the logic
- `DraggableElement` for the movable root element

This should allow us to use the dragging logic via events without loading the parts used to move the root element.

This change led me to be tempted to move the dragging logic to a decorator, which could add a `dragged` hook.

We could then easily use dragging capabilities in components without having to use the `Draggable` component as a child:

```js
import Base from '@studiometa/js-toolkit';
import withDragService from '@studiometa/js-toolkit/decorators/withDragService';

class Component extends withDragService(Base) {
  static config = {
    name: 'Component',
  };

  dragged({ x }) {
    this.x = x;
  }
  
  ticked() {
    this.$el.style.transform = `translateX(${this.x}px)`;
  }
}
```

The following props could be used:

- `isGrabbing` (`boolean`): whether the user is currently grabbing the targeted element or not (useful to watch for inertia)
- `x` (`number`): the current horizontal position in the viewport
- `y` (`number`): the current vertical position in the viewport
- `delta` (`{ x: number, y: number }`): the delta between the current position and the last
- `origin` (`{ x: number, y: number }`): the initial position where the dragging started
- `distance` (`{ x: number, y: number }`): the distance from the current position to the origin
- `target` (`{ x: number, y: number }`): the final position that will be reached at the end of the inertia

What are your thoughts on the subject @studiometa/js?

titouanmathis:
I missed an important thing on the service topic: the draggable logic need to known the target element to bind mouse/touch event listeners to trigger the beginning and the end of the drag. 

The service would need to have an option parameter to specify the targeted element:

```js
const { add } = useDrag();
const element = document.querySelector('.draggable');

add('key', (props) => {
  props.target.style.transform = `translate(${props.x}px, ${props.y}px)`;
}, { target: element });
```

And the decorator could specify the element to use in its options too:

```js
// Target the root element `$el`
class Component extends withDrag(Base, { target: '$el' }) {}

// Target the ref element `draggable`
class Component extends withDrag(Base, { target: '$refs.draggable' }) {}

// Use any value on the instance as target
class Component extends withDrag(Base, { target: (instance) => instance.$el }) {}
```

perruche:
> The basis of the drag logic and the root element displacement are now in 2 separate components:
> 
>     * `Draggable` for the logic
> 
>     * `DraggableElement` for the movable root element
> 
> 
> This should allow us to use the dragging logic via events without loading the parts used to move the root element.
> 
> This change led me to be tempted to move the dragging logic to a decorator, which could add a `dragged` hook.
> 
> We could then easily use dragging capabilities in components without having to use the `Draggable` component as a child:
> 
> ```js
> import Base from '@studiometa/js-toolkit';
> import withDragService from '@studiometa/js-toolkit/decorators/withDragService';
> 
> class Component extends withDragService(Base) {
>   static config = {
>     name: 'Component',
>   };
> 
>   dragged({ x }) {
>     this.x = x;
>   }
>   
>   ticked() {
>     this.$el.style.transform = `translateX(${this.x}px)`;
>   }
> }
> ```
> 
> The following props could be used:
> 
>     * `isGrabbing` (`boolean`): whether the user is currently grabbing the targeted element or not (useful to watch for inertia)
> 
>     * `x` (`number`): the current horizontal position in the viewport
> 
>     * `y` (`number`): the current vertical position in the viewport
> 
>     * `delta` (`{ x: number, y: number }`): the delta between the current position and the last
> 
>     * `origin` (`{ x: number, y: number }`): the initial position where the dragging started
> 
>     * `distance` (`{ x: number, y: number }`): the distance from the current position to the origin
> 
>     * `target` (`{ x: number, y: number }`): the final position that will be reached at the end of the inertia
> 
> 
> What are your thoughts on the subject @studiometa/js?

This is indeed a great idea to have both an abstract decorator to implement custom drag behavior and a ready to use component.
Maybe we should identify what behavior we want to have for the ready to use component, like nice default value for easing / inertia, etc. This might be more necessary for the next step for the slider component.
The documentation for the `withDrag` decorators will probably need some in depth advice on how to implement a performant componant

Issue #55, opened 2020-12-11
Add a decorator to mount Vue application inside a component


titouanmathis:
This would help stay consistent when mounting a Vue application inside of a Base component.

```js
import Base from '@studiometa/js-toolkit';
import withVue from '@studiometa/js-toolkit/decorators/withVue';
import VueComponent from './VueComponent.vue';

class ComponentWithVue extends withVue(Base) {
  static vueConfig = {
    components: {
      VueComponent,
    },
    render: (h) => h('VueComponent'),
  }

  get vueRootElement() {
    return this.$refs.vue;
  }

  mounted() {
     // Vue instance is available with `this.$vue`
    console.log(this.$vue);
  }
}
```

Vue should be added as an optionalDependency to the package.

The decorator should: 
- mount the vue app on the given root element when the component is mounted
- detroy the vue app when the component is destroyed
- update the vue app when the component is updated

PR #132, opened 2021-08-09
[Feature] Add a decorator to mount Vue application inside a component

LucasUlmer1:
This PR is related to [#55](https://github.com/studiometa/js-toolkit/issues/55).

It adds a decorator tied to a given component. This decorator mounts, updates and destroys a Vue app (on a given ref) when these hooks are called in the component.


packages/js-toolkit/decorators/withVue.js
```diff
@@ -0,0 +1,65 @@
...
+
+      if (!this.$vueConfig.components) {
+        throw new Error('[withVue] You must define a `component` key in vueConfig.');
+      }
+
+      if (typeof this.$vueConfig.components !== 'object') {
+        throw new Error('[withVue] `component` must be an object.');
+      }
+
+      if (!this.$vueConfig.render) {
```
titouanmathis, line 42: **suggestion:** Combine the 2 checks of the `render` property into one

```js
if (typeof vueConfig.render !== 'function') {
  throw new Error('[withVue] You must define a `render` function in vueConfig.');
}
```

packages/demo/src/js/components/MyVueComponent.js
```diff
@@ -0,0 +1,17 @@
+import Base from '@studiometa/js-toolkit';
+import { withVue } from '@studiometa/js-toolkit/decorators';
+import CustomComponent from './CustomComponent.vue';
+
+export default class MyVueComponent extends withVue(Base) {
+  static vueConfig = {
+    components: {
+      CustomComponent,
+    },
+    render: (h) => h('CustomComponent'),
```
titouanmathis, line 10: **suggestion:** We can directly use the imported component here without having to define it as a component

```js
static vueConfig = {
  render: (h) => h(CustomComponent),
}
```

packages/demo/package.json
```diff
@@ -17,5 +17,10 @@
     "prettier": "^2.3.0",
     "stylelint": "^13.13.1",
     "tailwindcss": "^2.1.2"
+  },
+  "optionalDependencies": {
+    "vue": "^2.6.14",
```
titouanmathis, line 22: **issue:** Vue should be a dependency as it is required for the demo to be compiled

I think we can drop the `vue-loader` and `vue-template-compiler` dependencies as they are already included in the `@studiometa/webpack-config` package.

titouanmathis (changes requested):
Some minor tweaks need to be done, but it looks promising, well done @LucasUlmer1!

Could you add a page for this decorator to the doc as well?

titouanmathis:
Closed in favor of #148.

PR #135, opened 2021-08-23
Add a `scrollTo(selector, options)` utility

titouanmathis:
- [ ] Add test
- [ ] Add doc
- [ ] Add demo


packages/js-toolkit/utils/scrollTo.js
```diff
@@ -0,0 +1,97 @@
...
+ * Handler for the click event on anchor links
+ *
+ * @param {string} selector The target selector.
+ * @param {Object} options Options for the scroll.
+ * @param {Object} options.offset Offset to add to the target scroll position.
+ * @param {Object} options.dampFactor The factor used for the scroll damping.
+ * @return {Promise<number>} A promising resolving with the target scroll position.
+ */
+export default function scrollTo(selector, options = { offset: 0, dampFactor: 0.2 }) {
+  const targetElement = document.querySelector(selector);
```
antoine4livre, line 13: It would be interesting to be able to pass a HTMLElement directly.

packages/js-toolkit/utils/scrollTo.js
```diff
@@ -3,14 +3,14 @@ import damp from './math/damp.js';
 /**
  * Handler for the click event on anchor links
  *
- * @param {string} selector The target selector.
+ * @param {HTMLElement} element The target element.
```
titouanmathis, line 6: **issue:** We need to accept both a selector or an element here

PR #140, opened 2021-08-30
[Bugfix] Improve types

titouanmathis:


packages/js-toolkit/components/Accordion/Accordion.js
```diff
@@ -39,7 +44,13 @@ export default class Accordion extends Base {
     name: 'Accordion',
     options: {
       autoclose: Boolean,
-      item: Object,
+      item: {
+        type: Object,
+        /**
+         * @return {AccordionItemStylesOption}
```
titouanmathis, line 50: **issue:** This should be `AccoprdionItemOptions`

packages/js-toolkit/components/Accordion/Accordion.js
```diff
@@ -50,7 +61,7 @@ export default class Accordion extends Base {
 
   /**
    * Init autoclose behavior on mounted.
-   * @this {Accordion & AccordionInterface}
+   * @this {AccordionInterface}
```
perruche, line 64: question: should this be `AccordionPrivateInterace` ?

titouanmathis: The `AccordionPrivateInterface` represents only the types we add ourselves, we need the `AccordionInterface` which merges both `Accordion` and `AccordionPrivateInterface` to be able to access the `Base` methods and properties.

PR #156, opened 2021-11-05
[Feature] Add a `withScrolledInView` decorator

titouanmathis:
This decorator can be used to easily setup animation based on scroll position of an element by adding a new `scrolledInView({ progress })` hook to any component, with `progress` being a value between `0` and `1` corresponding to the position of the root element in the viewport: 

- `0` when the root element enters the viewport from the bottom
- `0.5` when it is centered vertically in the viewport
- `1` when it exits the viewport from the top

```js
import { Base, withScrolledInView } from '@studiometa/js-toolkit';

class Component extends withScrolledInView(Base) {
  static config = {
    name: 'Component',
  };

  scrolledInView({ progress }) {
    this.$el.style.opacity = progress;
  }
}
```

See the [updated documentation](https://deploy-preview-156--studiometa-js-toolkit.netlify.app/api/decorators/withScrolledInView.html) and the [`Parallax`](https://github.com/studiometa/js-toolkit/blob/feature%2Fscrolled-in-view-decorator/packages/demo/src/js/components/Parallax.js) component in the demo to learn how to use it.

---

### Added
- Add a `withScrolledInView` decorator (#156)


titouanmathis:
- [ ] Add x axis values for horizontal scroll

packages/docs/api/decorators/withScrolledInView.md
```diff
@@ -0,0 +1,67 @@
...
+
+  scrolledInView({ progress }) {
+    const y = map(progress, 0, 1, -100, 100) * this.$options.speed;
+    this.$refs.target.style.transform = `translateY(${y}px)`;
+  }
+}
+```
+
+:::tip Info
+Apply transformations to refs instead of the root element. The decorator uses the [`getBoudingClientRect`](https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect) function to calculate the props values and it does not reverse any transform applied.
```
jeremschelb, line 66: `getBoudingClientRect` => `getBoundingClientRect`

PR #163, opened 2021-11-19
Add the proper link to config.options doc

notjb:
This PR Adds a proper link to an other page of the documentation.
✌️

### DIFF
--- a/packages/docs/api/instance-properties.md
+++ b/packages/docs/api/instance-properties.md
@@ -2,7 +2,7 @@
 
 ## `$options`
 
-An object containing the full options of the instance as defined in the [`config.options`](#options) property. Additionnally to the options defined in the config, the following properties are also available:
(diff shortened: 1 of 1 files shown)


packages/docs/api/instance-properties.md
```diff
@@ -2,7 +2,7 @@
 
 ## `$options`
 
-An object containing the full options of the instance as defined in the [`config.options`](#options) property. Additionnally to the options defined in the config, the following properties are also available:
+An object containing the full options of the instance as defined in the [`config.options`](api/configuration.html#config-options) property. Additionnally to the options defined in the config, the following properties are also available:
```
titouanmathis, line 5: **issue:** Use an absolute URL instead of a relative one

And I think we can increase the clickable area for the link in order to make it more visible 🙂 

```suggestion
An object containing the full options of the instance as defined in [the `config.options` property](/api/configuration.html#config-options). Additionnally to the options defined in the config, the following properties are also available:
```

titouanmathis (changes requested):

PR #168, opened 2021-11-29
Add managing refs section

KeziahMoselle:
There is still rewording to do but it should be good enough. Feel free to propose things to make it better!

### DIFF
--- a/packages/docs/guide/introduction/managing-refs.md
+++ b/packages/docs/guide/introduction/managing-refs.md
@@ -2,10 +2,98 @@
 
 ## What are refs?
 
+A ref is a *reference* to a DOM element. It allow us to quickly access HTML elements. We can define single or multiple refs that we can easily use in our component without manually querying DOM elements.
+
+> No more `querySelector` and `querySelectorAll`!
+
 ## Defining refs
+
+The [`refs` property](/api/configuration.html#config-refs) of the static `config` object should be used to define what refs will be used in the component otherwise it can't be accessed. This property requires an array of string that correspond to the refs names in the HTML code.
+
+We have 2 type of refs:
+- Single ref (example: `btn`)
+- Multiple refs (example: `items[]`)
+
+If you want to get an array of refs you can append `[]` to the ref name.
+
+```js {4}
+class Component extends Base {
+  static config = {
+    name: 'Component',
+    refs: ['btn', 'items[]'] // Declare the refs we will use
+    // `btn` is a single ref.
+    // `items[]` contains multiple refs and is always an array.
+  };
+}
+```
 
 ### Single refs
 
+First we need to declare the refs we want to use in our component, here we want to use a `btn`.
+
+```js {6,10}
+import { Base } from '@studiometa/js-toolkit';
+
+class Component extends Base {
+  static config = {
+    name: 'Component',
+    refs: ['btn'] // Declare the ref
+  };
+}
+```
+
 ### Multiple refs
 
+If we want to iterate on a ref we can append `[]` after the ref name. For example here `data-ref="items[]"` will yield an array of HTML elements.
+
+> Appending `[]` to a ref name will force the ref to an array even if there is only one occurence.
+
+```js {4,8}
+class Component extends Base {
+  static config = {
+    name: 'Component',
+    refs: ['items[]'],
+  };
+
+  mounted() {
+    this.$refs.items; // [<li data-ref="items[]">#1</li>, ...]
+  }
+}
+```
+
 ## Using refs
+
+Now that we defined the refs we want to use we can add them to our HTML code:
+
+
+```html {2,5-7}
+<div data-component="Component">
+  <button data-ref="btn">Click me</button>
+
+  <ul>
+    <li data-ref="items[]">#1</li>
+    <li data-ref="items[]">#2</li>
+    <li data-ref="items[]">#3</li>
+  </ul>
+</div>
+```
+
+They can be accessed directly in your instance with `this.$refs.<name>`
+
+```js {8-9}
+class Component extends Base {
+  static config = {
+    name: 'Component',
+    refs: ['btn', 'items[]'],
+  };
+
+  mounted() {
+    this.$refs.btn; // <button data-ref="btn">Click me</button>
+    this.$refs.items; // [<li data-ref="items[]">#1</li>, ...]
+  }
+}
+```
+
+::: tip Attaching events to a ref
+Visit the [Working with events](/guide/introduction/working-with-events.html) section of the guide to learn how to handle events while using refs.
+:::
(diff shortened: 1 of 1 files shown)


packages/docs/guide/introduction/managing-refs.md
```diff
@@ -2,10 +2,98 @@
 
 ## What are refs?
 
+A ref is a *reference* to a DOM element. It allow us to quickly access HTML elements. We can define single or multiple refs that we can easily use in our component without manually querying DOM elements.
```
perruche, line 5: **suggestion**: Avoid using "we", "us", "i" in the documentation as they are quite unsuitable in formal writing

**reword exemple**:
A ref is a *reference* to a DOM element. It allows to quickly access HTML elements. Define single or multiple refs to easily use them in a component without manually querying DOM elements.

packages/docs/guide/introduction/managing-refs.md
```diff
@@ -2,10 +2,98 @@
...
+- Multiple refs (example: `items[]`)
+
+If you want to get an array of refs you can append `[]` to the ref name.
+
+```js {4}
+class Component extends Base {
+  static config = {
+    name: 'Component',
+    refs: ['btn', 'items[]'] // Declare the refs we will use
+    // `btn` is a single ref.
```
perruche, line 24: **suggestion**: provide the return type instead of phrasing ?

packages/docs/guide/introduction/managing-refs.md
```diff
@@ -2,10 +2,98 @@
...
+    // `items[]` contains multiple refs and is always an array.
+  };
+}
+```
+
 ### Single refs
 
+First we need to declare the refs we want to use in our component, here we want to use a `btn`.
```
perruche, line 32: **personnal opinion**: The sections `single refs` and `multiple refs` seems repetitive since it's already explain above in `Defining refs`. Should defining refs be about where to register the refs in JS and in HTML ?

packages/docs/guide/introduction/managing-refs.md
```diff
@@ -2,10 +2,98 @@
...
+```js {4}
+class Component extends Base {
+  static config = {
+    name: 'Component',
+    refs: ['btn', 'items[]'] // Declare the refs we will use
+    // `btn` is a single ref.
+    // `items[]` contains multiple refs and is always an array.
+  };
+}
+```
```
perruche, line 28: **issue** There is no mention on the link between JS and HTML refs, should the exemple contain both HTML and JS code ? 
Maybe also add that if a ref is not define in the config it won't be available

packages/docs/guide/introduction/managing-refs.md
```diff
@@ -2,10 +2,98 @@
...
+```js {8-9}
+class Component extends Base {
+  static config = {
+    name: 'Component',
+    refs: ['btn', 'items[]'],
+  };
+
+  mounted() {
+    this.$refs.btn; // <button data-ref="btn">Click me</button>
+    this.$refs.items; // [<li data-ref="items[]">#1</li>, ...]
```
perruche, line 92: **question**: Isn't this is returning the Component instance and not the HTML element (which is in `this.$refs.btn.$el`) ?

perruche (commented):
Great addition.
Mostly nitpicking wording syntax to be consistant

packages/docs/guide/introduction/managing-refs.md
titouanmathis: The v2 no longer resolves component instances when a ref is also a component.

```html
<div data-component="Parent">
  <div data-ref="item" data-component="Child"></div>
</div>
```
```js
class Parent extends Base {
  static config = {
    name: 'Parent',
    refs: ['item'],
    components: { Child },
  };

  mounted() {
    console.log(this.$refs); // v1 → { item: Child }
    console.log(this.$refs); // v2 → { item: <div data-ref="item" data-component="Child"></div> }
  }
}
```

This was changed in order for refs to be more predictable about what they are. The `$children` property must be used to access child components' instances.

packages/docs/guide/introduction/managing-refs.md
titouanmathis: I think that the comments here can be misleading, we could improve it by displaying each ref on a single line:

```js
    refs: [
      'btn', // single ref
      'items[]', // multiple ref with the `[]` suffix
    ],
```

packages/docs/guide/introduction/managing-refs.md
```diff
@@ -2,10 +2,98 @@
...
+    name: 'Component',
+    refs: ['items[]'],
+  };
+
+  mounted() {
+    this.$refs.items; // [<li data-ref="items[]">#1</li>, ...]
+  }
+}
+```
+
```
titouanmathis, line 63: **suggestion:** add a small section about the preferred naming convention for multiple refs

- camelCase
- plural

packages/docs/guide/introduction/managing-refs.md
```diff
@@ -2,10 +2,98 @@
 
 ## What are refs?
 
+A ref is a *reference* to a DOM element. It allow us to quickly access HTML elements. We can define single or multiple refs that we can easily use in our component without manually querying DOM elements.
+
+> No more `querySelector` and `querySelectorAll`!
+
 ## Defining refs
 
+The [`refs` property](/api/configuration.html#config-refs) of the static `config` object should be used to define what refs will be used in the component otherwise it can't be accessed. This property requires an array of string that correspond to the refs names in the HTML code.
```
titouanmathis, line 11: **detail:** Change `will be used` with `will be available` 
**detail:** Remove unnecessary precision

```suggestion
The [`refs` property](/api/configuration.html#config-refs) of the static `config` object should be used to define what refs will be available in the component. This property requires an array of string that correspond to the refs names in the HTML code.
```

titouanmathis (changes requested):
This is a good first draft, well done @KeziahMoselle!

I have some things in mind that should be added, but they are not yet well thought. I will get back to you later on the subject.

packages/docs/guide/introduction/managing-refs.md
perruche: i confused myself, this is talking about refs and not components, of course it return HTLMElement 🤪

packages/docs/guide/introduction/managing-refs.md
KeziahMoselle: Should I also include the return type @titouanmathis ?

titouanmathis: I think this can be misleading here with an overload of information (maybe?) and we have the refs type described in the following examples, so I would advise not to include it.

KeziahMoselle:
I added back HTML examples in the Single ref and Multiple refs section @titouanmathis as it was not clear enough. What do you think of this version?

titouanmathis (approved):
👍

PR #170, opened 2021-12-06
Add managing options content

KeziahMoselle:
First draft of the managing options section of the guide.

### DIFF
--- a/packages/docs/guide/introduction/managing-options.md
+++ b/packages/docs/guide/introduction/managing-options.md
@@ -2,13 +2,59 @@
 
 ## What are options?
 
-An option is a way to pass data to a component.
+An option is a way to pass data to a component from the HTML template.
+
+Options are also a way to configure a component's behaviour at the instance level, since they are specific to each instance.
+
+```js {12-13}
+class Foo extends Base {
+  static config = {
+    name: 'Foo',
+    options: {
+      isActive: Boolean,
+    },
+  };
+
+  isOpen = false;
+
+  mounted() {
+    this.isOpen; // ← instance level but shared by all instances
+    this.$options.isActive; // ← instance level, can differ for each instance depending on the HTML
+  }
+}
+```
 
 ## Defining options
 
 The [`options` property](/api/configuration.html#config-options) of the static `config` object should be used to define what options will be available in the component.
 
 It is possible to define the type and the default value of the option, so that if your component expects a `number` it will automatically be parsed as a `number`.
+
+### Available types
+
+- `Boolean`: `data-option-bool` and `data-option-no-bool`, defaults to `false`
+- `Number`: parsed as number, defaults to `0`
+- `String`: parsed as a string, defaults to `''`
+- `Array`: parsed as an array, defaults to `[]`, default value must be defined with a function
+- `Object`: parsed as an object, defaults to `{}`, default value must be defined with a function
+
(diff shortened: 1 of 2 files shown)


packages/docs/guide/introduction/managing-options.md
```diff
@@ -1,7 +1,85 @@
...
+    name: 'VideoPlayer',
+    options: {
+      videoId: {
+        type: String,
+        default: 'dQw4w9WgXcQ'
+      },
+    },
+  };
+}
+```
```
titouanmathis, line 27: **suggestion:** List all available types and their usage

- `Boolean`: `data-option-bool` and `data-option-no-bool`, defaults to `false`
- `Number`: parsed as number, defaults to `0`
- `String`: parsed as a string, defaults to `''`
- `Array`: parsed as an array, defaults to `[]`, default value must be defined with a function
- `Object`: parsed as an object, defaults to `{}`, default value must be defined with a function

packages/docs/guide/introduction/managing-options.md
```diff
@@ -1,7 +1,85 @@
...
+It is possible to define the type and the default value of the option, so that if your component expects a `number` it will automatically be parsed as a `number`.
+
+```js {6-11}
+import { Base } from '@studiometa/js-toolkit';
+
+class VideoPlayer extends Base {
+  static config = {
+    name: 'VideoPlayer',
+    options: {
+      videoId: {
```
titouanmathis, line 20: **suggestion:** Add a mention about simple option definition (without default value)

```
options: {
  videoId: String,
  muted: Boolean,
}
```

packages/docs/guide/introduction/managing-options.md
```diff
@@ -1,7 +1,85 @@
...
+  mounted() {
+    this.$options.videoId; // 'dQw4w9WgXcQ'
+  }
+}
+```
+
 ## Updating options
+
+It is possible to update an option by reassigning its value:
+
```
titouanmathis, line 66: **suggestion:** Add a note to specify that reassigning an option is not recommended, as they should be used as initialization parameters.

For example, an internal flag `isActive` should not be an option if it is not meant to be changed in via the HTML API.

packages/docs/guide/introduction/managing-options.md
```diff
@@ -1,7 +1,85 @@
 # Managing options
 
+## What are options?
+
+An option is a way to pass data to a component.
```
titouanmathis, line 5: **suggestion:** Specify that it is a way to pass data from the HTML part of a component. 

Options are also a way to configure a component's behaviour at the instance level, since they are specific to each instance. 

```js
class Foo extends Base {
  static config = {
    name: 'Foo',
    options: {
      isActive: Boolean,
    },
  };

  isOpen = false;

  mounted() {
    this.isOpen; // ← instance level but shared by all instances
    this.$options.isActive; // ← instance level, can differ for each instance depending on the HTML
  }
}
```

titouanmathis (changes requested):
A good first version, thanks for your input @KeziahMoselle 👍

KeziahMoselle:
@titouanmathis I made some changes, tell me what you think. I think it still needs some work

titouanmathis (approved):
Great work @KeziahMoselle! 👍

PR #206, opened 2022-02-15
[Feature] Add support for multiple instance on a single element

titouanmathis:
This PR add support for multiple component instanciation on a single element, opening the road for more versatile components. 

A `WeakMap` is used to store the instances with the class used as key:  

```js

class App extends Base {
  static config = {
    name: 'App',
  };
}

class OtherApp extends Base {
  static config = {
    name: 'OtherApp',
  };
}

const app = new App(document.body).$mount();
const otherApp = new OtherApp(document.body).$mount();

console.log(app === document.body.__base__.get(App)); // true
console.log(app === document.body.__base__.get(OtherApp)); // false
console.log(otherApp === document.body.__base__.get(OtherApp)); // true
console.log(otherApp === document.body.__base__.get(App)); // false
```

This should not break existing components, as the `__base__` property should not be used outside of this package. 

This means, for example, that an anchor tag can be tied to 2 components to interact with 2 different parents, or we can create multiple small components to manage smooth scroll, Ajax navigation, etc. and bind them all to anchor tags on the page.

## To do

- [x] Update the doc with the new mechanism
- [x] Add example to the docs
- [x] Fix types


perruche:
🤟  oh yeah ! Really digging this !

notjb (commented):
🤯 
fun * App instances.
Curious about the use cases though 🤓

titouanmathis:
I am wondering if allowing to define multiple components in the `config.components` property would be interesting or not:

```js
class App extends Base {
  static config = {
    name: 'App',
    components: {
      a: [AnchorPrefetch, AnchorScrollto],
    },
  };
}
```

Maybe a `withMultipleComponents` decorator could handle this instead of adding the logic to the core:  

```js
class App extends Base {
  static config = {
    name: 'App',
    components: {
      a: withMultipleComponents(Base, [AnchorPrefetch, AnchorScrollto]),
    },
  };
}
```

PR #220, opened 2022-03-28
[Feature] Get a service props

titouanmathis:
Add support for accessing a registered service props easily from within a component:  

```js
import { Base,  } from '@studiometa/js-toolkit';

class Btn extends Base {
  static config = {
    name: 'Btn',
  };

  onClick() {
    if (this.$services.get('resized').breakpoint === 'xxs') {
      // Do something on small screens...
    }
  }
}
```

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -63,7 +63,7 @@ jobs:
 
           PRERELEASE=false
           # Check release type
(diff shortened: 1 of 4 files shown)


perruche:
Nice addition !

titouanmathis:
> Nice addition !

We were always reimporting services to access their props, lots of work for something that was already available in the `ServicesManager` 🙂

notjb (commented):
definitely will use this ^^

PR #225, opened 2022-04-04
[Feature] Add support for responsive options declaration

titouanmathis:
The aim of this PR is to add support for responsive options declaration with the `data-option-...` attributes. The suggested API is the following:

1. Declare an option with responsive capabilities in its configuration:

```js
class Component extends Base {
  static config = {
    options: {
      foo: {
        type: String,
        responsive: true
      },
    },
  };
}
```

2. Use `data-option-...` attributes to define different values for different breakpoints in your HTML:

```html
<div data-component="Component" 
  data-option-foo="defaultValue"
  data-option-foo[xs,s]="value for `xs` and `s` breakpoints"
  data-option-foo[l]="value for `l` breakpoint">
  ...
</div>
```

To be noted:

- This feature will not be compatible with `withFreezedOptions` decorator 
- This does not make options reactive, the different values are resolved when the option is accessed 

To do:

- [ ] Move the changes in `OptionsManager` to a child class, which would override the `get()` method (to avoid bloating)
- [ ] Find a way to change the managers used by a component to allow usage of child classes (dependency injection via a new `managers` parameter for the `constructor` method?)
- [ ] Use the `ResponsiveOptionsManager` class in the `withResponsiveOptions` decorator

---

@studiometa/core, does this seem like a **very useful**, **somehow useful** or **not needed** feature?


perruche:
Not sure this is important to us, but `data-option-foo[xs,s]="..."` does not pass W3C validator
I do link the syntax 😄

titouanmathis:
> Not sure this is important to us, but `data-option-foo[xs,s]="..."` does not pass W3C validator I do link the syntax 😄

That's why responsive options are readonly, an error is thrown when we do:

```js
element.dataset['data-option-foo[xs]'] = 'foo';
```

The `:` is valid, so we could use it instead:

```js
element.dataset['data-option-foo:xs'] = 'foo';
```

But the `,` is not valid, so we cannot use multiple breakpoints. Maybe we should use `min-width` tests instead of exact match?

titouanmathis:
I updated the syntax to use `:` as separator to be W3C compliant :

```diff
  <div data-component="Component" 
    data-option-foo="default value"
-   data-option-foo[xs,s]="value for `xs` and `s` breakpoints"
+   data-option-foo:xs:s="value for `xs` and `s` breakpoints"
-   data-option-foo[l]="value for `l` breakpoint">
+   data-option-foo:l="value for `l` breakpoint">
    ...
  </div>
```

cc @perruche

perruche:
> I updated the syntax to use `:` as separator to be W3C compliant :
> 
> ```diff
>   <div data-component="Component" 
>     data-option-foo="default value"
> -   data-option-foo[xs,s]="value for `xs` and `s` breakpoints"
> +   data-option-foo:xs:s="value for `xs` and `s` breakpoints"
> -   data-option-foo[l]="value for `l` breakpoint">
> +   data-option-foo:l="value for `l` breakpoint">
>     ...
>   </div>
> ```
> 
> cc @perruche

nice !

PR #252, opened 2022-07-14
[Release] 2.3.0

### LINKED ISSUE #248: Add support for native events on child components
For now, event bindings with `onChild...` methods will only work for events that are defined in the child component configuration. It would be interesting to be able to directly use a `onChildClick` method without having to define the `click` event in the `config.emits` array. We could then use child components more easily and without having to pass on native events.

titouanmathis:
## Added
- Add a `tween` utility function (#249)
- Add support for native event methods on child components (fix #248, #247, 3b6026d)
- Add `is...` utility functions: `isObject`, `isBoolean`, `isString`, `isNumber`, `isArray` and `hasWindow` (#247)
- Add an export of the `isDev` constant (#247)

## Changed
- Limit errors and warnings to dev environements (presence of a `__DEV__` global var) (#247)
- Refactor `is...` and `has...` test functions (#247)
- Change build target to `ES2022` (fa4ebd6)

## Fixed
- Add missing support for access to the `event.detail` data when working with `CustomEvent` with the event hooks (#250)
- Only bind event methods to the root element for defined or native events (#247, 0e202c4)


notjb:
👏

PR #257, opened 2022-07-18
[Feature] Refactor all services to be used via decorators

titouanmathis:
This PR refactors all services to be used on-demand via decorators. Its main goal is to reduce the initial size of the `Base` import.

⚠️ This is a breaking change.

## To do 
- [ ] Add tests 
- [ ] Update docs
- [ ] Add changelog


titouanmathis:
Closing for now as performance gains are small compared to the change in DX.

Issue #278, opened 2022-09-05
Bug with undefined ref in EventsManager `manageRef` function


titouanmathis:
The `elements` parameter can be `[undefined]`, throwing an error when trying to remove an event listener.

PR #279, opened 2022-09-05
[Bugfix] Filter the elements of the references in the EventsManager

w-jerome:
This PR filters the elements of the references in the `EventManager`, so that it does not attach events to elements of the DOM that no longer exist (#278)


titouanmathis (approved):
Thanks @w-jerome! 👍

PR #306, opened 2022-10-31
Fix drag service on touch devices

KeziahMoselle:
## Bug

On mobile the `drop()` function was never called on touch devices on Chrome mobile when dragging.

## The fix

Add `touchend` event support


titouanmathis (approved):
👍

PR #313, opened 2022-11-07
[Feature] Add import helpers and decorators for media queries

KeziahMoselle:
This PR adds a way to work with media queries.
A common use case would be to import a component dedicated to animations only if the user accepts motion. 

### `importOnMediaQuery(importFn, media)` Import helper

Usage:

```js
import { importOnMediaQuery } from '@studiometa/js-toolkit';

// ...

static config = {
    name: 'App',
    components: {
      AnimateComponent: () =>
        importOnMediaQuery(
          () => import('./components/AnimateScrollTest'),
          'not (prefers-reduced-motion)', // The media you want to match, here it mounts only if user accepts motion
        ),
    }
}
```

### `withMountOnMediaQuery(BaseClass, media)` Decorator

Usage:

```js
import { withMountOnMediaQuery } from '@studiometa/js-toolkit';

export default class MediaQueryDemo extends withMountOnMediaQuery(
  Base,
  'not (prefers-reduced-motion)', // The media query you want to match, if true the component will mount
) {
  static config = {
    name: 'MediaQueryDemo',
    log: true,
  };

  mounted() {
    this.$log('mounted');
  }

  destroyed() {
    this.$log('destroyed');
  }
}
```

You can also set the media query via the `data-option-media` attribute like so:

```twig
<div
  data-component="MediaQueryDemo"
  data-option-media="not (prefers-reduced-motion)">
</div>
```

### `importWhenPrefersMotion(importFn)`

```js
AnimateComponent: () =>
  importWhenPrefersMotion(
    () => import('./components/AnimateScrollTest')
  ),
```

### `withMountWhenPrefersMotion(BaseClass)`

```js
export default class AnimateComponent extends withMountWhenPrefersMotion(Base) {
  static config = {
    name: 'AnimateComponent',
  };
}
```
---

### Checklist

- [x] Add tests
- [x] Update documentation https://deploy-preview-313--studiometa-js-toolkit.netlify.app


KeziahMoselle:
I don't think adding dedicated functions for `prefers-reduced-data` is worth yet considering the compatibility table: https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-data#browser_compatibility

packages/js-toolkit/helpers/importOnMediaQuery.ts
```diff
@@ -0,0 +1,34 @@
...
+ *   The media query name and value (see https://developer.mozilla.org/en-US/docs/Web/CSS/@media#media_features)
+ * @returns {Promise<T>}
+ */
+export default function importOnMediaQuery<T extends BaseConstructor = BaseConstructor>(
+  fn: () => Promise<T | { default: T }>,
+  media: string,
+): Promise<T> {
+  let ResolvedClass;
+
+  const resolver = (resolve) => {
```
titouanmathis, line 19: **nitpick:** Could you refactor the resolver management which is the same for all import helpers?

I am thinkink of a `getResolver(fn)` function in the `helpers/utils.ts` file which could be used as follow:

```ts
import { getComponentResolver } from './utils.ts';

const resolver = getComponentResolver(fn);

return new Promise((resolve) => resolver(resolve));
```

packages/js-toolkit/helpers/importOnMediaQuery.ts
```diff
@@ -0,0 +1,34 @@
...
+    fn().then((module) => {
+      ResolvedClass = 'default' in module ? module.default : module;
+      resolve(ResolvedClass);
+    });
+  };
+  return new Promise((resolve) => {
+    const { matches } = window.matchMedia(media);
```
titouanmathis, line 26: **issue:** We need to listen to the change event to trigger the resolver if `matches` is falsy, as it can become truthy following a user interaction (resizing for example)

packages/js-toolkit/helpers/importWhenPrefersMotion.ts
```diff
@@ -0,0 +1,30 @@
+import type { BaseConstructor } from '../Base/index.js';
+
+/**
+ * Import a component if user does not reduce motion.
+ *
+ * @template {BaseConstructor} T
+ * @param {() => Promise<T|{default:T}>} fn
+ * @returns {Promise<T>}
+ */
+export default function importWhenPrefersMotion<T extends BaseConstructor = BaseConstructor>(
```
titouanmathis, line 10: **chore:** Use `importOnMediaQuery` here 

```js
export default function importWhenPrefersMotion(fn) {
  return importOnMediaQuery(fn, 'not (prefers-reduced-motion)');
}
```

packages/js-toolkit/decorators/withMountOnMediaQuery.ts
```diff
@@ -0,0 +1,115 @@
...
+     */
+    constructor(element: HTMLElement) {
+      super(element);
+
+      if (!this.$options.media && !media) {
+        this.$log('Missing either "$options.media" or "media" argument in withMountOnMediaQuery decorator.');
+        return;
+      }
+
+      this.__mediaQueryList = window.matchMedia(this.$options.media || media);
```
titouanmathis, line 72: **nitpick:** This variable does not need to be attached to the instance 

It is only used in the constructor, we can avoid adding another property to the class.

titouanmathis (changes requested):
Really nice PR, thanks @KeziahMoselle!

packages/docs/api/decorators/withIntersectionObserver.md
```diff
@@ -105,7 +105,7 @@ export default class Component extends withIntersectionObserver(Base, {
 }
 ```
 
-Or at the instance level in the `data-option-intersection-observer` attribute of the rool element:
+Or at the instance level in the `data-option-intersection-observer` attribute of the root element:
```
titouanmathis, line 108: **praise:** Nice catch!

packages/js-toolkit/decorators/withMountOnMediaQuery.ts
```diff
@@ -0,0 +1,107 @@
...
+
+    /**
+     * Listen for media query changes when the class in instantiated.
+     *
+     * @param {HTMLElement} element The component's root element.
+     */
+    constructor(element: HTMLElement) {
+      super(element);
+
+      if (!this.$options.media && !media) {
```
titouanmathis, line 57: **nitpick:** Add the `isDev` condition before logging anything

packages/js-toolkit/decorators/withMountOnMediaQuery.ts
```diff
@@ -0,0 +1,107 @@
...
+        this.$log(
```
titouanmathis, line 58: **suggestion:** Use the `this.$warn()` method

packages/js-toolkit/decorators/withMountOnMediaQuery.ts
```diff
@@ -0,0 +1,107 @@
...
+        setTimeout(() => this.$destroy());
+      }
+
+      const changeHandler = (event: MediaQueryListEvent) => {
+        this.__matches = event.matches;
+
+        if (event.matches) {
+          this.$mount();
+        } else {
+          setTimeout(() => this.$destroy());
```
titouanmathis, line 81: **nitpick:** This line should be tested

See my comment in the spec file for an example on how to test it.

packages/tests/decorators/withMountOnMediaQuery.spec.js
```diff
@@ -0,0 +1,43 @@
...
+  });
+
+  afterEach(() => {
+    matchMedia.clear();
+  });
+
+  it('should mount the component when user prefers motion', () => {
+    matchMedia.useMediaQuery(mediaQuery);
+    const instance = mountComponent();
+    expect(instance.$isMounted).toBe(true);
```
titouanmathis, line 35: **suggestion:** To test the missing line from the coverage with the `this.$destroy()` call

```suggestion
    expect(instance.$isMounted).toBe(true);
    matchMedia.useMediaQuery('(prefers-reduced-motion)');
    await wait(0);
    expect(instance.$isMounted).toBe(false);
```

packages/tests/decorators/withMountOnMediaQuery.spec.js
```diff
@@ -0,0 +1,43 @@
...
+  });
+
+  it('should not mount the component when user prefers reduced motion', () => {
+    matchMedia.useMediaQuery('(prefers-reduced-motion)');
+    const instance = mountComponent();
+    expect(instance.$isMounted).toBe(false);
```
titouanmathis, line 41: **suggestion:** To make sure that a component can be mounted again after a change in a media query, it might be interesting to add a few lines here

```suggestion
    expect(instance.$isMounted).toBe(false);
    matchMedia.useMediaQuery(mediaQuery);
    await wait(0);
    expect(instance.$isMounted).toBe(true);
```

titouanmathis (changes requested):
Some nitpicking to finalize this PR, but great work anyway! 👍

packages/js-toolkit/decorators/withMountOnMediaQuery.ts
KeziahMoselle: Isn't this log conditional already with the `log` option in the Class definition?

titouanmathis: It is indeed conditioned by the `this.$options.log` value. The `isDev` condition is different: the log will only be logged when in dev or watch mode, but the whole lines will be dropped (considered as dead code) in the final bundle when building for production.

This strategy is used for other debug and warning logs.

KeziahMoselle: Oh okay, that's very good to know! Adding this

titouanmathis (changes requested):

titouanmathis (approved):

titouanmathis (changes requested):
@KeziahMoselle could you update the changelog as well?

PR #319, opened 2022-12-02
[Feature] Refactor scheduler to use class

titouanmathis:


titouanmathis:
Not needed at the moment.

PR #323, opened 2022-12-02
[Feature] Resize screens definition

### LINKED ISSUE #264: Add support for defining an application's breakpoints without relying on reading them from the DOM
Currently, breakpoints values are read from the pseudo-elements `::before` and `::after` of a `[data-breakpoint]` element in the DOM. This works well, but is a little bit too magical for an easy use. 

In addition, the current `breakpoint` and `breakpoints` properties of the resize service props does not let us know if a given breakpoint is still active, forcing us to always test multiple breakpoints: 

```js
if (['l', 'xl', 'xxl'].includes(this.$services('resized').breakpoint)) {
  // ...
}
```

We could change this without a breaking change with 2 small additions:

- Add support for passing screens definition to the `createApp` function
- Add an `activeBreakpoints: Record<keyof breakpoints, boolean>` object to the resize service props 

**Example**

```js
import { screens } from '@studiometa/tailwind-config';
import { Base, createApp } from '@studiometa/js-toolkit';

class App extends Base {
  static config = {
    name: 'App',
  };
}

export default createApp(
  App,
  { screens },
);
```

With the `screens` export being: 

```js
export const screens = {
  xxs: '0px',
  xs: '480px',
  s: '768px',
  m: '1024px',
  l: '1280px',
  xl: '1440px',
  xxl: '1920px',
};
```

The `activeBreakpoints` property would be:

```js
const { activeBreakpoints } = useResize().props();

console.log(activeBreakpoints); 
{
  xxs: true,
  xs: true,
  s: true,
  m: false,
  l: false,
  xl: false,
  xxl: false,
}
```

titouanmathis:
This PR adds support for defining screens definition with the `createApp` helper.

Fix #264.

## To do 

- [x] Finalize feature
- [x] Add tests
- [x] Add docs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,14 @@ All notable changes to this project will be documented in this file. The format 
 
 ## [Unreleased]
 
+### Added
+
+- Add support for defining breakpoints when using the resize service ([#323](https://github.com/studiometa/js-toolkit/pull/323))
+
+### Changed
+
+- ⚠️ Refactor definition of breakpoints for the resize service ([#323](https://github.com/studiometa/js-toolkit/pull/323))
+
 ### Removed
 
 - ⚠️ Remove the `withVue2` decorator ([#395](https://github.com/studiometa/js-toolkit/pull/395))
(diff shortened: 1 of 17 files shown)


antoine4livre (approved):
Nice !

PR #330, opened 2022-12-12
[Feature] Add support for multiple component declaration in `data-component` attributes

titouanmathis:
Since [v2.0.0-beta.18](https://github.com/studiometa/js-toolkit/releases/tag/2.0.0-beta.18) (see #206), an HTML element can be used as root for multiple components. It was not possible yet to easily register multiple components on one element without using CSS selectors as component names. 

This feature adds support for registration of multiple components on a single HTML element via its `data-component` attribute:

```html
<div data-component="One Two Three"></div>
```

## To do
- [x] Write docs
- [x] Decide if the `,` as separator is the best option

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,7 +25,7 @@ yarn-error.log*
 # Build files
 dist/
 demo/dist/
-docs/.vuepress/dist/
+/packages/docs/.vitepress/cache
 coverage/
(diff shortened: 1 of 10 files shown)


titouanmathis:
Ping @jverneaut 😉

PR #348, opened 2023-02-03
[Bugfix] Listen for drop events on the window

jverneaut:
Listen for drop events on the window to prevent the target object from sticking to the mouse when the drop occurs outside the window boundaries.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file. The format 
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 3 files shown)


titouanmathis (changes requested):
Thanks for the fix 👍 

Some adjustments before merging:

- [x] Fix the tests by changing `div.dispatchEvent()` to `window.dispatchEvent()` in `drag.spec.js` line 46
- [ ] Update the `Unreleased` section of the changelog following the previous entries

titouanmathis (approved):
👍

PR #361, opened 2023-04-07
[Bugfix] Fix refs name normalization when used with prefix

jeremschelb:
- [x] Fix the bug.
- [x] Fix tests.
- [x] Create a test for the to event relative to ref with prefix.
- [x] Update changelog
- [x] Apply feedback

### DIFF
--- a/packages/js-toolkit/Base/managers/RefsManager.ts
+++ b/packages/js-toolkit/Base/managers/RefsManager.ts
@@ -2,22 +2,29 @@ import { AbstractManager } from './AbstractManager.js';
 import { isDev, isArray, isDefined, getAncestorWhereUntil } from '../../utils/index.js';
 
 const NORMALIZE_REF_NAME_REGEX = /\[\]$/;
+const NAMESPACED_REF_REGEX = /^(.*\.)/;
+const normalizedRefNamesCache = new Map();
 
 /**
  * Normalize the name of ref.
  */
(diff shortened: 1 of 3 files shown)


packages/js-toolkit/Base/managers/RefsManager.ts
```diff
@@ -7,7 +7,17 @@ const NORMALIZE_REF_NAME_REGEX = /\[\]$/;
  * Normalize the name of ref.
  */
 export function normalizeRefName(name: string) {
-  return name.endsWith('[]') ? name.replace(NORMALIZE_REF_NAME_REGEX, '') : name;
```
titouanmathis, line 10: **suggestion:** Add a cache to improve performance

The `normalizeRefName` function will be called multiple times with potentially the same value for the `name` argument. Adding a cache would improve performance by allowing us to execute the normalization only once for each value. 

```js
const normalizedRefNamesCache = new Map();

export function normalizeRefName(name: string) {
  if (normalizedRefNamesCache.has(name)) {
    return normalizedRefNamesCache.get(name)
  }

  // ...

  normalizedRefNamesCache.set(name, normalizedName);
  return normalizedName;
}
```

See the benchmark: https://jsbench.me/s0lg7t9k51.

titouanmathis (changes requested):
Nice PR 👍

titouanmathis (approved):

PR #411, opened 2024-02-16
[Feature] Migrate tests from Jest to Bun

titouanmathis:
For speed ⚡ 

## To-do 

- [x] Mock scrollHeight and scrollWidth
- [x] Mock history API
- [x] Reset DOM state after tests (tests run together fail, but they pass when run individually)


titouanmathis:
Need to wait for https://github.com/oven-sh/bun/issues/4015 to re-enable Codecov coverage upload.

PR #418, opened 2024-03-20
[Feature] `scrollTo` enhancement

### LINKED ISSUE #173: Improve `scrollTo` utility
- [ ] Allow giving a custom element to use as the scroll container, with default to `document.scrollingElement`
- [ ] Use `element.scrollTop` and `element.scrollLeft` to set scroll position

### LINKED ISSUE #261: Improve support of `scroll-behavior: smooth` with the `scrollTo` utility
When using the `scrollTo` utility with `scroll-behavior: smooth` applied on the scrolling element, the 2 animations are conflicting, resulting in laggy scroll in Chrome and Safari.

See https://codepen.io/titouanmathis/pen/ExEmZgb.

Maybe `scrollTo` should try to animate the scroll when `scroll-behavior` is set to `smooth`?

antoine4livre:
- [x] Directly use tween in the initial scrollTo function after the merge of the Tween refactor #419
- [x] Fix #173
- [x] Fix #261

### DIFF
--- a/packages/js-toolkit/utils/scrollTo.ts
+++ b/packages/js-toolkit/utils/scrollTo.ts
@@ -1,5 +1,6 @@
-import { TweenOptions, tween } from './tween.js';
-import { isString, isNumber, isObject, isFunction } from './is.js';
+import { tween } from './tween.js';
+import type { TweenOptions } from './tween.js';
+import { isString, isNumber, isFunction } from './is.js';
 import { lerp } from './math/index.js';
 
 export type ScrollPosition = {
--- a/packages/js-toolkit/utils/scrollTo.ts
+++ b/packages/js-toolkit/utils/scrollTo.ts
@@ -1,16 +1,16 @@
 import { tween } from './tween.js';
 import type { TweenOptions } from './tween.js';
-import { isString, isNumber, isFunction } from './is.js';
+import { isString, isNumber, isFunction, isDefined } from './is.js';
 import { lerp } from './math/index.js';
 
-export type ScrollPosition = {
+export interface ScrollPosition {
   left: number;
   top: number;
-};
+}
 
 export type ScrollTarget = string | HTMLElement | number | Partial<ScrollPosition>;
 
-export type ScrollToOptions = TweenOptions & {
+export interface ScrollToOptions extends TweenOptions {
   /**
    * Root element that will be scrolled.
    */
@@ -23,7 +23,7 @@ export type ScrollToOptions = TweenOptions & {
    * Distance from the target.
    */
   offset?: number;
-};
+}
 
 const xAxis = Symbol(0);
 const yAxis = Symbol(1);
@@ -98,7 +98,7 @@ function getTargetScrollPosition(
   top -= offset;
 
   if (axis !== bothAxis) {
-    if (axis !== xAxis) {
+    if (axis !== xAxis && !isDefined((target as Partial<ScrollPosition>).left)) {
       left = initialScrollPosition.left;
     }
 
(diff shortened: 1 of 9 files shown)


packages/js-toolkit/utils/scrollTo/tweenScrollTo.ts
```diff
@@ -0,0 +1,69 @@
...
+
+      window.removeEventListener('wheel', eventHandler);
+      window.removeEventListener('touchmove', eventHandler);
+
+      resolve(window.scrollY);
+    };
+
+    tw = tween(
+      (progress) => {
+        window.scrollTo({ top: lerp(window.scrollY, scrollTarget, progress) });
```
titouanmathis, line 50: **issue :** You need to save the value of `window.scrollY` before the tween starts

Otherwise, it will always be the latest value and the scroll animation might behave strangely.

titouanmathis (changes requested):

antoine4livre:
@titouanmathis We need to talk about the next modifications like #173 & #261.

What do you think about this new implementation ?

packages/js-toolkit/utils/scrollTo.ts
```diff
@@ -1,99 +1,216 @@
...
+import { lerp } from './math/index.js';
+
+// Set the type of ScrollPosition.
+type ScrollPosition = {
+  left?: number;
+  top?: number;
+};
+
+// Set the type of TweenScrollToOptions options.
+export type TweenScrollToOptions = TweenOptions & {
```
titouanmathis, line 12: **nitpick:** Rename the type to `ScrollToOptions`

packages/js-toolkit/utils/scrollTo.ts
```diff
@@ -1,99 +1,216 @@
...
+      if (rootElement) {
+        rootElement.removeEventListener('wheel', eventHandler);
+        rootElement.removeEventListener('touchmove', eventHandler);
+      } else {
+        window.removeEventListener('wheel', eventHandler);
+        window.removeEventListener('touchmove', eventHandler);
+      }
+
+      // Resolve the promise if the user interact with wheel or touchmove.
+      resolve(getCurrentScrollPosition());
```
titouanmathis, line 160: **suggestion:** refactor this part in a `stop` function 

It is used both in the `eventHandler` function and the `onFinish` hook.

packages/tests/utils/scrollTo.spec.ts
```diff
@@ -40,26 +34,32 @@ describe('The `scrollTo` function', () => {
 
   it('should scroll to a selector', async () => {
     expect(fn).not.toHaveBeenCalled();
-    useFakeTimers();
-    scrollTo('div');
-    runAllTimers();
-    useRealTimers();
-    expect(fn).toHaveBeenLastCalledWith({ top: 5000 });
+    await scrollTo('div');
```
titouanmathis, line 37: **suggestion :** Keep the use of the fake timers to speed up test execution

```js
useFakeTimers();
scrollTo('div', { duration: 1 });
await advanceTimersByTimeAsync(1016);
expect(fn).toHaveBeenLastCalledWith({ left: 0, top: 5000 });
useRealTimers();
```

titouanmathis (changes requested):

packages/js-toolkit/utils/scrollTo.ts
antoine4livre: ScrollToOptions is already the type of options for `window.scrollTo`, is it ok ?

packages/tests/utils/scrollTo.spec.ts
antoine4livre: Tween don't use timers right ? I can't make this fake timers work.

PR #478, opened 2024-07-05
[Feature] Add support for defining an option default value with a function

titouanmathis:
### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation, readme or JSdoc annotations)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality, like performance)
- [x] ✨ New feature (a non-breaking change that adds functionality)
- [ ] 🧹 Chore (updates to the build process or auxiliary tools and libraries)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

This PR adds support for defining an option default value with a function for all types of options. The function is called with the current instance as only parameter, allowing for advanced management of default values. 

```js {9-18}

class Player extends Base {
  static config = {
    name: 'Player',
    options: {
      type: {
        type: String,
        default(player) {
          switch (player.$el.constructor) {
            case HTMLVideoElement:
              return 'video';
            case HTMLAudioElement:
              return 'audio';
            default:
              return 'unknown';
          }
        },
      },
    },
  };
}
```

### 📝 Checklist

- [ ] I have linked an issue or discussion.
- [x] I have added tests (if possible).
- [x] I have updated the documentation accordingly.
- [x] I have updated the changelog.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,13 +4,14 @@ All notable changes to this project will be documented in this file. The format 
 
 ## [Unreleased]
 
+### Added
+
+- Add support for defining an option default value with a function ([#478](https://github.com/studiometa/js-toolkit/pull/478))
+- Add shorthand props on the scroll service for easier destructuring ([#432](https://github.com/studiometa/js-toolkit/pull/432))
+
 ### Fixed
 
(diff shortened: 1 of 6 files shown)


perruche:
stylé !

Issue #751, opened 2026-08-03
Explore declarative per-instance mount strategies with `data-mount`


titouanmathis:
## Context

While designing a browser CDN and DOM autoloader for [`@studiometa/ui`](https://github.com/studiometa/ui), we considered separating two declarative concerns:

```html
<div
  data-component="MapboxMap"
  data-load="visible"
  data-mount="visible">
</div>
```

- `data-load` would be owned by the CDN loader and decide when the component module is imported and globally registered.
- `data-mount` would be owned by js-toolkit and decide when each concrete component instance mounts.

Import timing is naturally class-wide: once one occurrence triggers an import, the constructor is globally registered. Mount timing needs to be per instance so that loading one visible `MapboxMap` does not immediately mount every offscreen map.

`data-mount` is deliberately out of scope for the first CDN implementation because it affects framework lifecycle semantics and needs a design that works beyond the CDN use case.

## Existing primitives

js-toolkit already provides class-level decorators such as:

- `withMountWhenInView`;
- `withMountOnMediaQuery`;
- `withMountWhenPrefersMotion`.

`withMountWhenInView` instantiates the component, mounts it when it enters the viewport, destroys it when it leaves, and mounts it again when it returns.

These decorators work when a component author chooses the lifecycle policy at class-definition time. They do not directly provide a per-element policy declared by markup.

## Why this cannot simply be implemented by the CDN

The global registry maps one component name to one constructor. Dynamically wrapping constructors according to each element's `data-mount` value would introduce identity conflicts:

- two occurrences of the same component could request different mount policies;
- `config.components` contains the original child constructors;
- parent-managed, globally registered, and async children must keep matching constructor identities;
- components already using a mount decorator could receive nested or conflicting observers;
- `$closest`, `$query`, instance lookup, events, teardown, and the global MutationObserver must continue to work consistently.

A CDN-specific wrapper would therefore risk creating a second lifecycle system or registering multiple constructors under the same canonical component name.

### Attribute semantics

How should a strategy apply when one element declares several components?

```html
<div data-component="Action Dialog Transition" data-mount="visible">
```

Possible interpretations:

1. Apply the strategy to every component on the element.
2. Support a per-component grammar.
3. Treat `data-mount` as invalid when more than one component is declared.
4. Introduce component-qualified attributes instead.

The design should also cover selector-based registrations and configured attribute prefixes.

### Visibility semantics

`visible` can mean two different things:

1. Mount once on first intersection and remain mounted.
[...]

Issue #651, opened 2025-08-01
[Base] Add support for multiple option type


titouanmathis:
Sometimes an option can be either a number or an array of numbers. Having the ability to specify both types could be useful:

```js
class Component extends Base {
  static config = {
    name: 'Component',
    options: {
      offset: [Number, Array],
    },
  };
}
```

The option could then be used with a number or an array:

```html
<div data-component="Component" 
  data-option-offset="10">
  ...
</div>

<div data-component="Component" 
  data-option-offset="[10, 20]">
  ...
</div>
```

Issue #630, opened 2025-05-12
[Feature] Enable bubbling and cancelation for custom events


titouanmathis:
Currently, custom events dispatched by the `Base` class do not bubble and are not cancelable. 

It might be interesting to enable both options, but it might be a breaking change.

Issue #627, opened 2025-05-12
[Base] Maybe improve `config` merge strategy?


titouanmathis:
Currently, the static `config` property of a class extending `Base` or another class extending `Base` is partially merged with its parent with the following strategy: 

- the config objects are merged: `config = { ...parent.config, ...config }`
- options are merged: `config.options = { ...parent.config.options, ...config.options }`
- emits are merged: `config.emits = { ...parent.emits, ...config.emits }`
- refs are not merged
- children components are not merged

See [Base.ts#L131-L158](https://github.com/studiometa/js-toolkit/blob/develop/packages/js-toolkit/Base/Base.ts#L131-L158) for the full strategy.

This means that a child class redeclaring some refs or some components will inherit from its parent options, but not its refs or components. This can lead to some errors when a parent class is assuming that a given ref or component is declared in its configuration. 

```js
import { Base } from '@studiometa/js-toolkit';

class Parent extends Base {
  static config = {
    name: 'Parent',
    refs: ['one'],
  };

  mounted() {
    this.$refs.one.textContent = 'Hello';
  }
}

class Child extends Parent {
  static config = {
    name: 'Child',
    refs: ['other'],
  }
}

const child = new Child(document.body);

// This will throw, as the `Child` refs config overrides the `Parent` one
await child.$mount();
```

## Current solution 

The solution when extending is to merge refs and components in the `Child` config when extending the `Parent`:

```diff
 class Child extends Parent {
   static config = {
     name: 'Child',
+    refs: ['other', ...Parent.config.refs],
+    components: { ...Parent.config.components },
   };
 }
```

The solution for component developers is to make sure the components and refs used in their components are defined before accessing them:

```diff 
 class Parent extends Base {
   static config = {
     name: 'Parent',
     refs: ['one'],
   };
 
   mounted() {
+    if (this.$refs.one) {
       this.$refs.one.textContent = 'Hello';
+    }
   }
 }
```

## Framework level solution

We could mitigate this kind of issue by implementing a solution at the `Base` class level. 

- Merging `refs` and `components` like `options` and `emits`
- Adding a config to indicate which config property needs to be merged or overriden
- Add a catch-all key to both `$refs` and `$children` objects to avoid triggering an error (with a proxy)

The solution should allow both package developers and consumers to not worry about defined or not defined components.

PR #759, opened 2026-08-11
Prototype the v4 architecture

titouanmathis:
## What

A working prototype of the v4 architecture discussed in #694, as a new private workspace package `packages/v4`, plus its design document, real-browser tests, a live demo, and an export-size comparison against the current package.

**Core model: the registry is the framework, and the DOM is the component tree.** An instance exists because its element is in the DOM and its class is registered. Nothing else creates or destroys instances.

## Contents

- **`packages/v4/DESIGN.md`** — the consolidated design: the five objectives (independent components, one registry, auto-mount on insertion/ejection, parents listening to child events, children advertising their existence), the destroy ≠ terminate ≠ disconnected lifecycle triad, `mounted()` returning its cleanup, the frame-aligned scheduler, native view transitions in core, and the provide/inject shared-state primitive. Supersedes the custom-elements and bespoke-CDN parts of the earlier spec draft.
- **`packages/v4/src/`** — zero-dependency TypeScript implementation (~1100 lines): record-based registry observer, `Base` with bubbling cancelable `$emit` + `on<Child><Event>` delegation + `$watchChildren` + `$provide`/`$inject`, the context protocol with reactive `Cell`s, the read/write/afterWrite/background scheduler with cancelable handles, and batched `viewTransition`.
- **`packages/v4/tests/`** — 13 tests running in headless Chromium through Vitest browser mode (`@vitest/browser-playwright`). Real-browser runs caught two bugs happy-dom missed: moved elements skipping their destroy (stale watchers on both ancestors) and the `$watchChildren` sweep firing during field initialization.
- **`packages/v4/demo/`** — Vite-served page (`npm run dev` in `packages/v4`) reimplementing ui components on the v4 model: Accordion on native `<details>`, a scroll-snap Slider whose `{ index, total }` lives in one provided cell (replacing the `connectChildren` handshake), TodoList with view transitions, and a `withMountWhenInView` decorator where the IntersectionObserver drives the reversible mount/destroy cycle (#751's `in-view` strategy).
- **`.github/workflows/export-size.yml`** — switched to `weareikko/export-size@1` (as in studiometa/ui), measuring both packages in one report.

## First numbers

| Export | v3 (min+gzip) | v4 prototype (min+gzip) |
| --- | ---: | ---: |
| barrel (`.`) | 26.63 kB | 3.38 kB |
| `Base` | 9.06 kB | 2.89 kB |
| `registerComponent` | — | 3.23 kB |

The prototype covers a fraction of v3's surface (no services, decorators, autoload manifests, responsive options), so this is not like-for-like — but the core model itself is an order of magnitude smaller.

## Related

#694, #630, #697, #751, #627, #651

https://claude.ai/code/session_01FqH6T9mNAPYGjQjj7Y9XmU


titouanmathis:
### On the CodSpeed check

It surfaced one real bug, which is fixed in d4f1c98: the comparison benchmark set the `blocking` feature at module scope, and since CodSpeed runs every benchmark file in a single process, that global setting followed the rest of the suite. Vitest isolates files into separate workers locally, which is why it only ever appeared in CI. Fixing it also exposed that vitest does not run `beforeAll` in benchmark mode — the hook version silently measured an undefined instance.

The check still reports a degradation, and I read the remainder as an environment mismatch rather than a regression:

- the benchmarks it flags **change identity between runs** — first `prototype (scoped DOM scan)` and `create with simple keyframes`, now `translate3d (x only)` and `scaleX, scaleY, scaleZ`;
- all of them are v3-only code paths this branch never touches;
- they all move by roughly the same ~18%, which reads as the whole HEAD run being measured on a different machine than BASE;
- CodSpeed says as much itself: *"Different runtime environments detected — some benchmarks with significant performance changes were compared across different runtime environments, which may affect the accuracy of the results."*

A uniform shift across arbitrary, unrelated benchmarks is not something a code change produces. I would acknowledge it in the CodSpeed UI rather than chase it further, but flagging it here so the check being red is not mistaken for a finding.

The numbers this PR actually adds are in `DESIGN.md` under **Measurements**, and were taken locally where both sides run in the same environment.

Issue #780, opened 2026-08-13
v4: remaining work before the ui 2.0 rewrite


titouanmathis:
Tracking issue for everything still outstanding on the v4 prototype.

**Where v4 stands** (reconciled 2026-08-18 against `main` at #856). Three rounds merged (#777, #778, #779), then the round this issue was opened for (#781 → #789), and then **54 more PRs, #790 → #846**, which this body had fallen behind. **Fifteen** ui families are ported as feasibility tests — `Accordion`, `Dialog`, `ScrollAnimation`, `Slider`, `Data*`, `Action`, `ClickOutside`, `InView`, `Track`, and from #847 `Fetch`, `LazyInclude`, `Prefetch`, `Cursor`, `Draggable`, `Carousel` — and the gap list they produce, now 1–43 in `migration/REPORT.md`, is what drove this roadmap. **That list is now closed apart from three entries**, and the work since has come from reviewing the closures rather than from porting anything new.

**Every section is closed.** F2 landed in #860, which closes this issue — the v4 prototype's outstanding work is done.

**Section E is closed by #847**, the last feasibility round: every family this roadmap named is ported, and the ten gaps it filed are section **K**. Sections **A**, **B**, **C**, **D**, **I** and **J** are closed. What is left is **F** (two ui-level refactors), **K** (the port round's open gaps) and three decisions in **G**.

**#847's review is where four of those gaps came from, and it closed three of them.** `swap()` grew the axis its consumers asked for, `@on` learned to type its own handler, `smoothTo()` became usable by the components it was written for, and two families stopped writing to the DOM in the scheduler's read phase. #848 followed from the same review: `interaction:page`.

**Then #849 → #856 closed the investigation round**, and three of those PRs exist because a question found something the roadmap had not: removing `$terminate()` uncovered a live listener leak, making booleans presence-only turned three port fixtures from quietly-right into plainly-wrong, and the first lint rule found a scheduler-lane defect in `ScrollAnimation` that no test could fail on.

Section **B is closed**: #785 landed every correctness item, two of them by deciding the reported behaviour was the contract rather than a bug. **D2**, **I5**, **I6**, **C8** and **C9** landed with it.

Sizes: **S** ≈ one focused change, **M** ≈ needs design judgement, **L** ≈ a round of its own.

[...]

PR #789, opened 2026-08-14
feat(v4): C9 — restore v3's namespaced data-ref form

titouanmathis:
Addresses **C9** of #780. Does not close it.

## The gap

`belongsTo()` returns false if any `data-component` element sits between a ref and the component root. That is the right default almost always, and it leaves no way at all to say *this ref belongs to the `Slider` further up*. v3 had the escape hatch — `data-ref="Slider.next"` — and v4 dropped it.

## What is restored

v3's semantics, from `RefsManager.refBelongToInstance()`:

- The plain form `data-ref="next"` stops at the nearest component of **any** kind. Unchanged.
- The namespaced form `data-ref="Slider.next"` walks past every boundary **except another component of that name**. So the nearest `Slider` wins, and a nested `Slider` still shadows its parent.
- The **root is never tested** against the namespace. It cannot be: ui registers `FigureShopify` under the name `Figure`, so `data-ref="FigureShopify.img"` sits under `data-component="Figure"`. The namespace names the class's `config.name`, which is what the selector already matched on; the walk only decides who else could have claimed it.
- The token form is honoured — `data-component="Slider Other"` is a `Slider` — via `selectorFor()`, where v3 compared the whole attribute. v4 supports several components per element and v3 did not.

Nothing downstream carries the namespace. `Slider.dots[]` is `$refs.dots`, `onDotsClick()` and `@on('dots[]', …)`.

## The `[]` interaction

The namespace wraps the whole definition, **suffix last**: `data-ref="Slider.dots[]"`.

That is v3's order — `RefsManager.__register()` builds `` `${name}.${refName}` `` from the already-suffixed config entry — and it is the order ui's own documentation is written in (`declarative-runtime.md`: `data-ref="Modal.close[]"`). It also reads correctly: `[]` qualifies the ref, and the namespace qualifies the pair. Tested, including the mixed case where one property is fed by both spellings and has to keep one document order for `on<Ref><Event>`'s `index`.

## The namespace lives in the markup, not in `config.refs`

This is the one design point worth reading, because it is the opposite of what was assumed.

`FigureShopify` declares `refs: ['img']` **once**, inherited from `AbstractFigure`, and its templates write `data-ref="img"` three times and `data-ref="FigureShopify.img"` once — chosen per template by how deeply the image is nested. A namespace in the declaration cannot express that: it would force one component to pick a single spelling for all of its templates, and it would make an inherited ref undeclarable, since `AbstractFigure` cannot know the name of the subclass that will use it.

So the two forms are **not** two spellings of one thing, the way the `[]` question was. They are two different questions — *who is my nearest owner* and *who is my named owner* — answered into one property. **One spelling still holds where it applies**: the declaration names the entry (`@on('dots[]', …)`), and the derived names drop what is derivable (`$refs.dots`, `onDotsClick`).

## The three ui usages

| file | attribute | what it does |
|---|---|---|
| `Frame/stories/form/app.twig:13` | `data-ref="App.form"` | reaches a `Frame`'s `<form>` from the app root, past `data-component="Frame"` |
| `FigureShopify/stories/reveal.liquid:16` | `data-ref="FigureShopify.img"` | reaches an `` wrapped in `data-component="Transition"` |
| `guide/concepts/declarative-runtime.md:103` | `data-ref="Modal.close[]"` | documents the form, and fixes the suffix order |

All three are the same shape: a ref wrapped in a presentational component. All three work under this change.

## Cost

`queryRefs()` runs once per ref per cache miss, so the second spelling was measured rather than assumed. Over a 25-element subtree in Chromium:

| | one selector | `[data-ref="a"],[data-ref="b"]` | two queries |
|---|---|---|---|
| 25 matches | 8.0 µs | 11.2 µs | 8.7 µs |
| 0 matches | 1.2 µs | 3.4 µs | 2.1 µs |

A selector list costs Chromium its single-attribute fast path, so this ships **two queries**: the plain one keeps its fast path, the namespaced one is a cheap second query that in practice returns nothing. `compareDocumentPosition` is paid only when a ref is genuinely written both ways under the same component.

`memo()` was considered for `refPropertyName()`, as v3 memoises its equivalent, and rejected: it is one `endsWith` and one `slice` on a short string, run once per ref per mount. Nothing to buy back.

## Tests

`packages/v4/src/Base.spec.ts`, five cases: a namespaced ref resolving across an intervening component (with the plain form still scoped as before, in the same markup), the suffix-last list spelling with both forms feeding one property in document order, a namespaced ref addressed to another component being ignored, a nested component of the same name shadowing its parent, and `on<Ref><Event>` delegating to a namespaced list ref under its unnamespaced method name.

Full v4 suite green (627 passed, 1 expected fail), `tsc --noEmit`, oxlint, `oxfmt --check` and `subpaths:check` green. The two `unicorn(no-useless-spread)` warnings in `packages/v4/src/context.ts:93` and `packages/js-toolkit/src/autoload/loader.ts:467` are pre-existing and untouched.

`origin/main` is merged in, so #783 is included. #787 also touches `Base.ts` (`@on` handler binding, ref-named handlers included), so a merge pass is expected there.

https://claude.ai/code/session_017MZV1ZifTfhQkHm4tavc5a


titouanmathis:
Merged `origin/main` at `611c63b6`, picking up #787 (C8). No textual conflict in `Base.ts`, and the two changes agree on inspection:

- `#bindHandlers()` stores the **declaration** as `entry.name` on both paths — `@on` matches `refs.find(({ definition }) => definition === child)` per #787, the magic-name path stores `ref.definition` — and delegation resolves it through `isRefOf(el, root, componentName, entry.name)`. So `@on` never sees a namespace, and a namespaced attribute reaches its handler through the plain declared name: `data-ref="Slider.next"` declared as `next` binds `@on('next', 'click')` and `onNextClick()`.
- `warnRefSuffixMismatch()` **cannot** fire on a namespaced attribute. It compares the decorator's string against `config.refs` and never reads the DOM, so the attribute spelling is not an input to it. No fix needed; `dddfad1c` pins it with a test asserting silence as well as delivery, so a later change making it DOM-aware would be caught.

`dddfad1c` also corrects the comment above `refs` in `#bindHandlers()`, which still described `@on()` as naming a ref by its derived spelling — already stale on `main` after #787, and stale on two counts now.

Re-verified after the merge: v4 suite 636 passed / 1 expected fail across 48 files, `lint:types`, oxlint, `oxfmt --check` and `subpaths:check` all green. The two `unicorn(no-useless-spread)` warnings remain pre-existing and untouched.

The `config.refs: ['Slider.next']` warning stays an unruled follow-up, not added.

PR #791, opened 2026-08-14
Investigation(v4): where the size goes, and what could be simplified

titouanmathis:
Refs #780.

An investigation, not a refactor. No runtime code changes — **two documents**, `packages/v4/SIZE.md` and `packages/v4/SIMPLIFY.md`.

---

# 1. `SIZE.md` — where the bytes are

"Under 10 kB gzipped" is ambiguous, and the three plausible readings differ by a factor of nine. The document names one and reports the other two beside it:

| Metric | Baseline |
| --- | --- |
| **A — one component** (`Base` + `registerComponent` + transitive graph, as a bundler ships them) | **9.14 kB gzip** |
| B — critical path (`registerManifest()` bootstrap) | 9.20 kB gzip bundled; 45.96 kB gzip / 15 requests / 5 RTT as published |
| C — the whole package | 16.44 kB gzip; `dist/` is 106 modules, 82.72 kB gzip |

**v4 is already under 10 kB on metric A, with 0.86 kB of headroom.** The work is keeping it there, not getting there.

**Top three reductions**

1. **Ship `dist/` minified** — CDN slice 45.96 → **13.23 kB gzip** (−71 %), same request count and RTT. `dist/` ships every JSDoc block today. Costs a bundler user nothing.
2. **Break the `registry.ts` → `Base.ts` static edge** — manifest bootstrap 9.20 → **3.89 kB gzip**, 15 → 11 requests, 5 → 4 RTT. `Base` is imported for one `instanceof`. Prototyped and validated against the full suite; metric A moves +0.01 kB.
3. **Move `Base`'s eleven optional methods off the class** — metric A 9.14 → **7.47 kB gzip** (−18 %). 41 % of the metric is one class and no bundler tree-shakes a method. Presented as a trade with its cost stated, and recommended *against* for now.

Stacking everything that trades no feature away reaches **6.80 kB**; the floor is **~5.2 kB**.

---

# 2. `SIMPLIFY.md` — what is over-engineered, and what only looks it

Run under one rule, because `SERVICES-REVIEW.md` records 17 confirmed defects each fixed with a regression test: **a branch is not over-engineering because it looks defensive.** Nothing was called unnecessary without checking whether a test fails without it, whether a comment or design doc explains it, and whether `git log -S` shows it arriving with a fix.

| Category | Count |
| --- | ---: |
| 1 — unjustified, remove | 10 |
| 2 — justified but undocumented, document | 19 |
| 3 — justified and documented, leave alone | ~115 |

The headline is the negative result. `Base.ts` carries 79 JSDoc blocks in 1 950 lines, most naming the naive alternative and why it fails — which is why category 3 dominates and the removal list is short.

**Top findings**

- **Six error-isolation guards around `Base`'s lifecycle hooks have no regression test.** Measured: deleting all eleven `try`/`catch` in `Base.ts` fails exactly two tests; deleting just those six leaves the suite green. They are justified — error isolation is a real property of this framework — but it is written down for the *scheduler* only. `SIZE.md` prices deleting them at 0.12 kB, so nothing currently stands in the way of someone doing it.
- **`ingest()`'s attribute filter is dead** — the observer's `attributeFilter` already guarantees its predicate. Git history shows it as a leftover from a superseded design; it can only lose work, never protect any.
- **`optionAttributes()` re-walks a chain `resolveConfig()` already caches** — two config keys have already shipped as bugs for exactly this shape.
- Removals also cover `perTarget()`'s unused `keyOf`, five zero-reference type aliases, `destroyWithin`'s unreachable fallback, `$off()`, and `test-utils.ts` being emitted into `dist/`.

**Three latent defects** surfaced while checking guards, reported separately from the audit and each proved with a throwaway probe spec (since deleted): one element can hold two `DragService` instances; an `<svg data-component>` mounts but never reconciles; a one-shot `$inject` silently drops the teardown its own documented contract promises.

**What I expected to find and did not:** a pile of dead public API. A naive unused-symbol sweep flags `registerManifest`, `until()` and `swap()` — all three are argued for in `DESIGN.md` and heavily tested. In a private prototype at `0.0.0`, "no consumer in this repository" is not evidence.

Also answered: both `SERVICES-SURFACE.md` findings left open on purpose are **still open** — only two commits touched `services/` since, and neither addresses them.

---

`SIZE.md` was amended where the audit corrected it (§R7: the config-chain cleanup needs no module move). Both documents cross-reference rather than repeat where they overlap.

https://claude.ai/code/session_017MZV1ZifTfhQkHm4tavc5a

### DIFF
--- a/packages/v4/SIMPLIFY.md
+++ b/packages/v4/SIMPLIFY.md
@@ -1,0 +1,709 @@
+# Simplify — what is over-engineered, and what only looks it (2026-08-14)
+
+Companion to `SIZE.md`. That document asked what v4 costs; this one asks what v4
+could stop doing. They are different questions with different answers, and the
+headline here is a negative result: **of everything that reads as excessive
+caution, almost none of it is. The work is documentation, not deletion.**
+
+Audited: every runtime module in `packages/v4/src`, plus what `migration/` and
+`demo/` actually consume. Method and its limits in §6.
+
+## The rule this audit was run under
+
+`SERVICES-REVIEW.md` records **17 confirmed defects** found by adversarial
+review, each fixed with a regression test. Much of what looks defensive here is
+the scar of a measured failure — `memo()` uses `has()` rather than an `undefined`
+check because the cheap version recomputes a memoised `undefined` forever; the
+services fan-out walks a snapshot because iterating the live set delivered props
+to subscribers that did not exist yet.
+
+So nothing was called unnecessary without checking three things: **does a test
+fail without it, does a comment or design doc explain it, and does `git log -S`
+show it arriving with a fix.** A yes to any one makes it justified, and the only
+question left is whether the reason is written down.
+
+Two disciplines follow, and they are what makes this list short:
+
+- Where a guard is called untested, that was established by **deleting it,
+  rebuilding and running the suite** — not by reading.
+- Where something is called unused, "no consumer in this repository" was treated
+  as **weak** evidence. v4 is a private prototype at version `0.0.0`; its
+  consumers do not exist yet, and `migration/` is ten ported families, not a user
+  base. §5 lists three exports a naive unused-symbol sweep flagged that are
+  plainly justified.
+
+---
+
+## Summary
+
+| Category                                  | Count |
+| ----------------------------------------- | ----: |
+| 1 — unjustified, remove                   |    10 |
+| 2 — justified but undocumented, document  |    19 |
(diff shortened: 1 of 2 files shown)


titouanmathis:
Closing this for now. It is too early to optimize or simplify v4 while its API and runtime design are still changing. We can repeat the measurements and audit after the feature set is stable.

PR #802, opened 2026-08-15
[Bugfix] Snapshot scheduler tick callbacks

titouanmathis:
### Type of change

- [x] Bug fix

### Description

Iterate a snapshot of scheduler tick callbacks for each frame. Check each callback against the live set before calling it. This makes subscriptions added during a tick start on the next frame, while an earlier unsubscribe still prevents a callback from running in the current tick.

Add regression coverage for callback addition, callback removal before its turn, and bounded repeated callback creation. Keep the scheduler benchmark variants aligned with the shipped implementation.

### Validation

- `npm run test -w @studiometa/js-toolkit-v4 -- src/scheduler.spec.ts src/services/raf.spec.ts` (26 passed)
- `npm run test:v4` (745 passed)
- `npm run lint:types`
- `npm run lint:static`
- `npm run lint:fmt`
- `npm run subpaths:check`
- `npm run build`
- `npm run bench -w @studiometa/js-toolkit-v4 -- src/scheduler.bench.ts`

The post-fix synthetic tick benchmarks remain below 0.00025 ms per frame for both the 1-subscriber and 100-subscriber class cases.

### Checklist

- [x] I have added tests.
- [ ] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.
- [ ] I have updated the changelog.

### DIFF
--- a/packages/v4/src/scheduler.bench.ts
+++ b/packages/v4/src/scheduler.bench.ts
@@ -241,7 +241,11 @@ class ClassScheduler implements Variant {
         delta: this.#clampDelta(frameTime),
       };
       this.#lastFrameTime = frameTime;
-      for (const callback of this.#tickCallbacks) {
+      const callbacks = [...this.#tickCallbacks];
+      for (const callback of callbacks) {
+        if (!this.#tickCallbacks.has(callback)) {
+          continue;
+        }
         try {
           callback(props);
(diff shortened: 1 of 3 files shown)


titouanmathis:
No changes were needed. The automated review found no blocking issues, and it created no review threads.

PR #803, opened 2026-08-15
[Bugfix] Make v4 service startup reentrant

titouanmathis:
### Type of change

- [x] Bug fix

### Description

Mark a `createService()` run as starting before calling its source. A subscription created by a synchronous startup emit now joins that run instead of starting and leaking a second source. Roll back subscriptions atomically when startup throws so the service can start again.

Add regression coverage for nested startup subscriptions, cleanup, restart, and startup-throw recovery.

### Test evidence

- `npm exec --workspace=@studiometa/js-toolkit-v4 -- vitest run src/services/service.spec.ts`
- `npm run test:v4`
- `npm run lint:types`
- `npm run lint:static`
- `npm run lint:fmt`
- `npm run subpaths:check`
- `npm run build`
- `npm run bench --workspace=@studiometa/js-toolkit-v4 -- --run src/services/service.bench.ts`

### Checklist

- [x] I have added tests.
- [x] No public documentation change is required.
- [x] No changelog change is required.

### DIFF
--- a/packages/v4/src/services/service.spec.ts
+++ b/packages/v4/src/services/service.spec.ts
@@ -50,6 +50,95 @@ describe('createService', () => {
     expect(calls).toEqual(['start', 'stop', 'start']);
   });
 
+  it('joins a subscription created by a synchronous startup emit to the same run', () => {
+    const calls: string[] = [];
+    const service = createService<number>({
+      props: () => 1,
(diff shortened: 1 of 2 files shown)


titouanmathis:
Bot review found no concrete defects. No code changes are needed.

PR #826, opened 2026-08-16
feat(v4): add createGroup, the missing half of v3's withGroup

titouanmathis:
## Verdict: one small primitive was missing, and it is not `$group`

Contexts cover **finding** the group. They do not cover **being the group**.

`provide`/`inject` + `subscribeContext()` already solve the hard half of v3's `withGroup`: nearest-provider-wins gives a member its closest group, a pending request is replayed when a provider appears later, and the mount announcement re-answers a member when a *nearer* group shows up. That is the whole of Disclosure's `DISCLOSURE_CONNECTED` / `DISCLOSURE_GROUP_CONNECTED` / `useMutation` / `__connect` / `__disconnect` dance, deleted.

What nothing in v4 provides is the **roster** on the coordinator's side:

- `$watchChildren` is the closest thing, and it is wrong here: it collects every matching descendant *across nested group boundaries*, so a nested `DisclosureGroup`'s items land in the outer group's collection. Disclosure has to filter them back out by asking each item which group it belongs to.
- A bare `Set` — what v3 handed out — has no value cell. A coordinator can read its peers but never learn that one arrived. With v4 mounting on DOM insertion and no ordering guarantee, an invariant over the set (*one open at a time*) has to be re-checked on every membership change, not established once.

So: **`createGroup()`** — 90 lines, one new runtime export, built on `signal()` and reached as a context value. No parallel registry; the report's argument against one stands and this does not touch it.

```ts
interface Group<T extends GroupMember = GroupMember> {
  /** The roster in document order, a new array on every change. */
  readonly members: Signal<readonly T[]>;
  /** Join, and get the leave function back. */
  join(member: T): () => void;
}
```

Two design points worth arguing about:

- **`join()` returns `leave`** because that is already the shape of `subscribeContext()`'s answer/teardown contract. A member's whole group wiring is one call, and migration to a nearer group leaves the old one *before* joining the new one for free.
- **Document order is the tie-breaker**, deliberately. The task asked what happens when a peer mounts later; the answer is that mount order never decides. Which peer keeps its state is a fact about the markup.

## The Disclosure pattern in v4

Whole thing, no decorators, no build step. This is `src/group.spec.ts` verbatim.

```js
const DisclosureGroupContext = createContext('DisclosureGroup');

class DisclosureGroup extends Base {
  static config = { name: 'DisclosureGroup', options: { multiple: Boolean } };

  #peers = createGroup();

  // Provided during construction, so a member that mounts first still resolves it.
  api = this.$provide(DisclosureGroupContext, {
    members: this.#peers.members,
    join: (peer) => this.#peers.join(peer),
    open: (peer) => this.open(peer),
  });

  get members() {
    return this.#peers.members.value;
  }

  /** A peer arriving or leaving can break the invariant, so re-check the roster. */
  mounted() {
    return this.#peers.members.subscribe(() => this.reconcile(), { immediate: true });
  }

  open(peer) {
    if (!this.$options.multiple) {
      for (const other of this.members) {
        if (other !== peer) other.setOpen(false);
      }
    }
    peer.setOpen(true);
  }

  /** Document order decides, so the outcome does not depend on mount order. */
  reconcile() {
    if (this.$options.multiple) return;
    const [, ...extra] = this.members.filter((peer) => peer.isOpen);
    for (const peer of extra) peer.setOpen(false);
  }
}

class Disclosure extends Base {
  static config = { name: 'Disclosure', refs: ['trigger', 'panel'], options: { open: Boolean } };

  group;
  isOpen = false;

  mounted() {
    this.setOpen(this.$options.open);
    // The nearest group, whenever it appears.
    return subscribeContext(this.$el, DisclosureGroupContext, (group) => {
      this.group = group;
      const leave = group.join(this);
      return () => {
        leave();
        this.group = undefined;
      };
    });
  }

  onTriggerClick() {
    this.toggle();
  }

  toggle() {
    if (this.isOpen) this.setOpen(false);
    else if (this.group) this.group.open(this);
    else this.setOpen(true);
  }

  setOpen(open) {
    this.isOpen = open;
    this.$refs.panel.hidden = !open;
    this.$refs.trigger.setAttribute('aria-expanded', String(open));
  }
}
```

Against v3's `Disclosure.ts` + `DisclosureGroup.ts` this drops: two document-level `CustomEvent` channels and their listeners, `useMutation(document, …)` with its per-instance key, `__connect` / `__disconnect` / `register` / `unregister`, `__scheduleReconcile` and its `nextTick` debounce, and the `items` getter's own DOM sort and `contains` filter. What is left is the invariant, which was always the only part that belonged to the group.

### What `group.spec.ts` asserts

13 tests in real Chromium. The ones that matter:

| case | outcome |
| --- | --- |
| open peer mounts **after** the group, later in the DOM | it loses; the earlier one keeps its state |
| open peer mounts **after** the group, earlier in the DOM | it wins; the later one closes |
| group's `data-component` added **after** its members mounted | it takes them all |
| nested group | takes its own members; the outer roster excludes them |
| inner group inserted around an already-mounted member | the member changes hands — **same instance**, asserted, so this is the `subscribeContext()` re-answer and not a destroy/remount |
| member's element removed | it leaves the roster |
| no group above | the disclosure still toggles on its own |

Removing the `members.subscribe()` reconcile fails exactly the two late-peer tests and nothing else, so the roster-as-a-value is load-bearing and not decoration.

## How the two `Data*` sites translate

They already are translated, in `packages/v4/migration/Data/` — and **they do not use `createGroup()`**. That is the interesting part, so spelling it out:

- **`DataScope`** (v3: `getScopedGroups(this)` + a `WeakMap` of scope → name → `Set`) becomes one field initializer: `registry = this.$provide(DataRegistryContext, new DataRegistry({ scoped: true, … }))`. The scope resolution, the `WeakMap` and the page-global fallback all collapse into nearest-provider-wins plus `provideRootContext()`.
- **`DataBind`** (v3: `withGroup(Base, 'data:', { getScope, getGroup })` and `this.$group`) becomes one `subscribeContext()` in `mounted()`: the callback joins `registry.join(group, this)`, the teardown leaves, and `mounted()` returns the unsubscribe. `getDataScope()`'s manual `__base__` ancestor walk is gone.

`DataRegistry` keeps its own membership `Set` rather than a `Group`, and that is the right call: **the two v3 consumers of `withGroup` divide on what "group" means.** `Disclosure` groups by *ancestry* — the group is a component in the DOM, and the whole difficulty was the two sides finding each other. `Data*` groups by *name*, with the nearest `DataScope` only choosing which partition table to read; membership is one field of a per-name record beside values, sources and hydration state, and nothing ever subscribes to a roster. It has the identical `join(group, member) → leave` shape for the identical reason. A group that is a set of peers gets `createGroup()`; a group that is a partition of a keyed store keeps the store. DESIGN.md §5 now records that line.

## What is deliberately not here

- **No `$group`.** Nothing implicit, nothing global, no name option. The name a v3 consumer passed was the partition key of a page-global registry; a group whose membership is a DOM fact needs neither a name nor a scope resolver.
- **No `Base` method.** `createGroup()` imports `signal` and nothing else — no component graph on the `./createGroup` subpath — and the member side is a `subscribeContext()` call that already has a home.
- **No disconnected-member sweep.** v3 swept on every read of `$group` because membership lived on a registry that outlived the instance. v4 destroys a component when its element leaves the DOM, so the member's own teardown removes it. Asserted.

## Verification

`npm run lint`, `npm run lint:types`, `npm run test:v4` (988 passed) and `npm run check:package` are green on each commit. Root runtime exports go 79 → 80; `packages/v4/src/subpaths/createGroup.ts` and the exports map are generated by `npm run subpaths`.

### DIFF
--- a/packages/v4/DESIGN.md
+++ b/packages/v4/DESIGN.md
@@ -614,6 +614,41 @@ The reactive container is called `Signal` — the name the ecosystem settled on 
 
 This is the property @studiometa/ui's `DataChannel` gets from alien-signals today, and the reason it can be dropped rather than depended on — its `publish()` always builds a fresh frame object, so the `===` bail-out never fires there and what it actually relies on is _one delivery per subscriber per settle, carrying the latest value_. Settling stays in the same task on purpose: `DataBind` echoes form-control input, and a microtask hop would be a visible change of behaviour. The price is that a subscriber writing unconditionally on every delivery live-locks the loop rather than overflowing the stack, which is the same trade every synchronous reactive graph makes.
 
+#### A set of peers — `createGroup()`
+
+v3's `withGroup` gave every instance a `$group: Set<Base>` of its peers, keyed by a group name and optionally scoped by a resolver. §5's `provideRootContext()` section states why it is not ported: the set had no value cell. What that section did not settle is where the _membership_ itself lives, and the answer is not a second registry. **A group is a membership published as one `Signal`, created by whoever coordinates it and reached the way every other shared value is reached — as a context.**
+
+```js
+// The coordinator owns the membership and hands out the ways in.
+const peers = createGroup();
+api = this.$provide(DisclosureGroupContext, {
+  members: peers.members, // the members to read, in document order
+  join: (peer) => peers.join(peer), // returns the leave function
+  open: (peer) => this.open(peer), // the invariant stays here
+});
+```
+
+```js
+// A member joins the nearest group, whenever that group appears.
+mounted() {
+  return subscribeContext(this.$el, DisclosureGroupContext, (group) => {
+    this.group = group;
+    const leave = group.join(this);
+    return () => { leave(); this.group = undefined; };
+  });
+}
+```
+
+`join()` returning its own `leave` is the whole ergonomic point: the shape of `subscribeContext()`'s answer/teardown contract already _is_ the shape of joining and leaving a group, so a member that migrates to a nearer group leaves the old one before it joins the new one, with no `__connect`/`__disconnect` handshake to write. Scoping comes free from nearest-provider-wins, so a nested group takes its own members and never its parent's — the case `$watchChildren` cannot express, since it collects every matching descendant across nested boundaries.
+
+**The membership is a value, and that is what v3's `Set` was missing.** A coordinator subscribes to it, so a peer arriving or leaving is an event it can act on. That matters because v4 mounts on DOM insertion with no ordering guarantee: "the set of my peers" is not settled at any point in time, and an invariant over the set — one open at a time, one selected item — has to be re-checked on every change rather than established once. **Document order is the tie-breaker, deliberately**, so which peer keeps its state is a fact about the markup and not about which one happened to mount first. A disclosure written open that mounts late loses to the one before it in the DOM, and wins over the ones after it; both orders are asserted in `group.spec.ts`, which builds the Disclosure pattern end to end from this helper plus `$provide` and `subscribeContext` alone.
+
+Nothing sweeps disconnected members, and nothing needs to: v4 destroys a component when its element leaves the DOM, so the member's own teardown is what removes it. v3 swept on every read of `$group` because its membership was written from `mounted` and `destroyed` on a registry that outlived both.
+
+The helper holds a `Set` and a `Signal` and nothing else — no element index, no name keys, no global map. `createGroup()` names no group and resolves no scope: the name a v3 consumer passed was the partition key of a page-global registry, and a group whose membership is a DOM fact needs neither.
+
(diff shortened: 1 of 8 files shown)


titouanmathis:
The `packages/v4/node_modules` finding is stale — it describes a push that has since been amended away.

Verified on the reviewed commit itself (`842427d1`, the current head):

```
gh pr diff 826 | grep -c node_modules            → 0
git ls-tree -r origin/feature/v4-grouping        → 0 matches
```

The PR's file set is eight files, all under `packages/v4/`: `DESIGN.md`, `package.json`, `exports.spec.ts`, `group.ts`, `group.spec.ts`, `index.ts`, `subpaths/createGroup.ts` and `test/package-node-consumer.js`.

The symlink was mine — I created it so the agent worktree could resolve dependencies, and an amend swept it into the commit. It was removed before this head was pushed. Merging.

PR #837, opened 2026-08-16
refactor(v4): reduce cyclomatic complexity at the top of the distribution

titouanmathis:
Six behavior-preserving refactors on `packages/v4/src`, each in its own commit, from a cyclomatic complexity analysis of the package.

## Why

Complexity in v4 is concentrated, not spread: before this PR, 8% of the functions held 31% of the total, and a single function held 30 of it. These commits take the top of that list. No public API changes, no behavior changes — every diagnostic code, message and ordering is preserved byte-for-byte, and the whole suite passes unchanged.

## Result

| Metric | Before | After |
| --- | --- | --- |
| Functions | 830 | 857 |
| Total complexity | 1890 | 1904 |
| Mean | 2.28 | 2.22 |
| **Max** | **30** | **18** |
| **Functions at CC ≥ 11** | **14** | **8** |

The total goes up by 14 because extraction adds function boundaries — expected, and not the metric that matters here. What matters is that the worst function drops from 30 to 3. Every remaining function at CC ≥ 11 is one this PR does not target.

## The commits

| Commit | Function | CC |
| --- | --- | --- |
| Split `processMutations` into its named phases | `registry.ts` | 30 → 3 |
| Route lifecycle failures through one guard helper | `Base.ts`, 10 call sites | −20 |
| Split `bindHandlers` into its three binding phases | `Base.ts` | 15 → 3 |
| Read option values from a per-type table | `Base.ts` | 12 → 1 |
| Split `conflictingKeys` into its two comparisons | `decorators.ts` | 13 → 1 |
| Isolate subscriber failures with one shared helper | `context.ts`, `services/service.ts` | −6 |

Two of them remove real duplication rather than only branches: ten copies of the same `try/catch` + `reportLifecycleFailure` block in `Base.ts`, and four copies of the subscriber-failure block across `context.ts` and `services/service.ts`.

## Cost

Bundled per export, minified and gzipped, against `main`:

| Export | main | This PR |
| :-- | --: | --: |
| `(barrel)` | 22.65 kB | +225 B (+1.0%) |
| `component` | 6.75 kB | +95 B (+1.4%) |
| `registerComponent` | 6.44 kB | +66 B (+1.0%) |
| `Base` | 8.05 kB | +128 B (+1.6%) |

## Not done, deliberately

- **A mount-strategy dispatch table.** It was written, measured and dropped: it replaced a short-circuiting if-chain with a `Map` lookup plus a per-call context object, on the path the registry runs for every element/component pair, and cost ~130 B on top. `applyMountStrategy` stays at CC 11.
- **`transform` (18), `createElement` (15), `reconcileElement` (14).** Above the rest of the distribution, but their complexity is breadth, not depth — splitting them would trade a flat table of cases for indirection.

## Notes for review

- `collectAttributeChanges` (`registry.ts`, CC 12) is the one function this PR creates above 11. It is a flat classification of attribute names into three buckets; splitting it further would add indirection without adding clarity.
- No `*.spec.ts` file was modified.

## Verification

- `npm run test:v4` — 80 files, 1083 tests passing, unchanged from `main`.
- `npm run lint` — clean.

### DIFF
--- a/.oxlintrc.json
+++ b/.oxlintrc.json
@@ -12,6 +12,12 @@
       {
         "allowDestructuring": true
       }
+    ],
+    "complexity": [
+      "error",
+      {
+        "max": 20
+      }
     ]
   },
   "overrides": [
@@ -53,8 +59,9 @@
       }
     },
     {
-      "files": ["**/*.spec.ts", "**/*.bench.ts", "**/__benchmarks__/**"],
+      "files": ["**/*.spec.ts", "**/*.bench.ts", "**/__benchmarks__/**", "packages/v4/test/**"],
       "rules": {
+        "complexity": "off",
         "js-toolkit/no-internal-barrel-import": "off",
         "js-toolkit/require-config": "off",
         "js-toolkit/require-config-name-pascal-case": "off",
--- a/packages/js-toolkit/src/utils/css/animate.ts
+++ b/packages/js-toolkit/src/utils/css/animate.ts
@@ -150,6 +150,9 @@ const TRANSLATE_DEFS: ReadonlyArray<readonly [string, number, string]> = [
  * Compile a pair of keyframes into a pre-computed segment.
  * All constant work (deltas, property filtering) is done here once.
  */
+// Deliberately flat: this runs once per segment so that rendering does not
+// branch, and splitting it would move the branches back onto the hot path.
+// oxlint-disable-next-line complexity
 function compileSegment(from: NormalizedKeyframe, to: NormalizedKeyframe): CompiledSegment {
   const transformStarts: number[] = [];
   const transformDeltas: number[] = [];
--- a/packages/v4/src/Base.ts
+++ b/packages/v4/src/Base.ts
@@ -693,6 +693,22 @@ const handlerPlan = /* @__PURE__ */ memo((ctor: BaseConstructor): HandlerPlan =>
   return { childNames, refs, componentName: config.name, entries };
 });
 
+/** One handler waiting on a descendant, resolved at event time. */
+type DelegatedEntry =
+  | { kind: 'child'; name: string; invoke: (payload: DelegatedEvent) => void }
+  // `name` here is the **declared** ref, suffix included: it is compared
+  // against the `data-ref` attribute and handed to `queryRefs()`.
+  | { kind: 'ref'; name: string; invoke: (payload: RefEvent) => void };
+
+type DelegatedEntries = Map<string, DelegatedEntry[]>;
+
+function addDelegated(delegated: DelegatedEntries, type: string, entry: DelegatedEntry): void {
+  if (!delegated.has(type)) {
+    delegated.set(type, []);
+  }
+  delegated.get(type)?.push(entry);
+}
+
 /**
  * `mounted()` may return one cleanup, several, nothing — or a promise of
  * those. The shape nests, so a subclass composes with what it extends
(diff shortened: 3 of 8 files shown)


titouanmathis:
Good catch — that one was a real regression, not measurement noise.

**Cause.** `mount-strategies.ts` declares `MOUNT_ATTRIBUTE` next to the strategy code, and on `main` the constant subpath shook down to just the string. The new dispatch table keyed its entries through two frozen constant objects (`MOUNT_STRATEGY_NAMES.eager`, `MOUNT_ARGUMENTS.forbidden`). A property read is not something a bundler may drop — it cannot prove the object has no getter — so the table survived tree-shaking, and with it `applyEager` and the `scheduler` and `diagnostics` modules behind it. Every bundle importing the constant alone paid for the whole strategy engine.

**Fix** (4ab4fce4): literal keys in the table, plus `/* @__PURE__ */` on the `new Map`. The two constant objects are gone — they were internal, used once each, and duplicated the `MountStrategy` type declared right above them. The named-constant convention is for values consumers spell, which these were not.

Measured locally with tsdown, minified and gzipped, same method for both:

| Export | main | before fix | after fix |
| :-- | --: | --: | --: |
| `MOUNT_ATTRIBUTE` | 80 B | 232 B | **80 B** |

Back to the baseline byte for byte.

**The rest of the diff is genuine.** With that fixed, the remaining growth is what the seven refactors actually cost — more function boundaries, the strategy table, the guard helper:

| Export | main | branch | Diff |
| :-- | --: | --: | --: |
| `(barrel)` | 22.65 kB | 23.00 kB | +354 B (+1.6%) |
| `component` | 6.75 kB | 7.00 kB | +248 B (+3.7%) |
| `registerComponent` | 6.44 kB | 6.66 kB | +219 B (+3.4%) |
| `Base` | 8.05 kB | 8.18 kB | +128 B (+1.6%) |

That is the trade this PR makes: roughly +1.6% on the barrel to cut the worst function from CC 30 to 3 and halve the number of functions above 11. The single most expensive item is the mount-strategy table, +129 B for CC 11 → 5 — the one worth arguing about if the bytes matter more than the branch count.

Tests unchanged: 80 files, 1083 passing.

titouanmathis:
Dropped the mount-strategy table — the if-chain is back, unchanged from `main`. Both commits that touched `mount-strategies.ts` are gone (the table itself and the tree-shaking fix it needed), so the branch is now six commits and the file does not appear in the diff at all. Force-pushed.

Reason: the table cost bytes *and* runtime. It replaced a short-circuiting if-chain with a `Map` lookup plus a per-call context object, on a path that runs for every element/component pair the registry schedules. Not a trade worth making for six points of branch count.

**Size, against `main`** — same tsdown/minify/gzip method throughout:

| Export | main | with table | now |
| :-- | --: | --: | --: |
| `(barrel)` | 22.65 kB | +354 B | **+225 B (+1.0%)** |
| `component` | 6.75 kB | +248 B | **+95 B (+1.4%)** |
| `registerComponent` | 6.44 kB | +219 B | **+66 B (+1.0%)** |
| `Base` | 8.05 kB | +128 B | **+128 B (+1.6%)** |
| `MOUNT_ATTRIBUTE` | 80 B | 232 B → 80 B | **80 B** |

**Complexity, against `main`:**

| Metric | Before | After |
| :-- | --: | --: |
| Max | 30 | **18** |
| Functions at CC ≥ 11 | 14 | **8** |
| Mean | 2.28 | **2.22** |

`applyMountStrategy` stays at 11, as it is on `main`. Everything else the PR set out to fix is still fixed: `processMutations` 30 → 3, `bindHandlers` 15 → 3, `conflictingKeys` 13 → 1, the option reader 12 → 1.

Tests: 80 files, 1083 passing. `npm run lint` clean.

titouanmathis:
Added the guardrail as a seventh commit (4fc72c12): `"complexity": ["error", { "max": 20 }]` in `.oxlintrc.json`, repository-wide.

20 sits above every function in the repository today, so it reports the *next* one rather than handing anyone a backlog — a ratchet, not a cleanup list. Scoping it to v4 was the alternative; repo-wide costs one exemption and covers v3 and the eslint plugin too.

**What it took to make it pass:**

- Test files, benchmarks and `packages/v4/test/**` are exempt. The v4 runtime harness has a CC-35 `run()`, but that is fixture branching, not logic. `**/*.spec.ts` and `**/*.bench.ts` were already grouped in an override; the harness directory joins them.
- One inline exemption, in v3: `compileSegment()` in `utils/css/animate.ts`, at 23. It is flat on purpose — it runs once per segment so that rendering does not branch — so the comment says that rather than pretending it is a to-do.

Nothing else in the repository needed touching. For reference, the next functions down are two eslint-plugin rules at 20, then `utils/css/transform.ts` at 19.

`npm run lint` is clean: oxlint exits 0, formatting and types pass.

PR #855, opened 2026-08-18
feat(eslint-plugin): add the v4 lint rules and a v4 config

titouanmathis:
A v4 consumer installs one config:

```js
import { jsToolkit } from '@studiometa/eslint-plugin-js-toolkit';

export default [jsToolkit.configs.v4];
```

Before this, there was nothing to install and nothing ran. `.oxlintrc.json` enabled the js-toolkit rules for `packages/js-toolkit/src` and `packages/demo` only, so `packages/v4` — the package these rules describe — got zero of them, and a consumer's only option was to list every rule by hand. The v3 set is also actively wrong for v4: `require-emit-declared-in-config` polices `config.emits`, which v4 removed. So `configs.v4` and the `packages/v4/**` override are not packaging polish; without them the rules run nowhere.

The plugin is loaded by oxlint as a JS plugin and exported as an ESLint flat config from one authoring shape, so all of this works in both runners with no new package. A JS plugin gets no type information — `--type-aware` drives oxlint's built-in `typescript/*` rules through tsgolint and hands no checker to JS plugins — so every rule here is type-free. That is verified, not assumed: the plugin's rules do run under `oxlint --type-aware packages/v4`.

## Rules

| Rule | Catches | Severity |
| --- | --- | --- |
| `no-write-in-read-phase` | A DOM write in a frame hook, a read-phase service subscriber, or a `$read()` task. The frame service's contract is that the callback **returns** the function which mutates. | error |
| `no-options-assignment` | `this.$options.x = …`, `Object.assign(this.$options, …)` and — since #854 made the view non-writable — `this.$options = …`. All three already throw; the rule moves the failure to CI. | error |
| `prefer-instance-scheduler` | `defaultScheduler.read()` / `.write()` inside a component, where `this.$read()` / `this.$write()` tie the task to the instance and cancel on destroy. | warn |
| `option-default-factory` | A literal object or array as an option `default`, which every instance then shares. Core warns at runtime, but only once a component mounts. | error |
| `no-conflicting-negated-option` | A component declaring both a boolean option `x` and an option named `noX`, making `data-option-no-x` mean two things. | error |
| `no-deprecated-properties` (extended) | With `{ version: 'v4' }`: `$update`, `$warn`, `$log`, `$terminate`, `$services.enable()`/`.disable()`, `updated()`, `terminated()`, `config.emits`, plus the existing three. | error |

### `no-write-in-read-phase`

The one that motivated the item. `useRaf()` fans its subscribers out inside `defaultScheduler.read()`, and two of the fifteen ported families broke the contract independently — `Cursor` wrote a transform straight from the frame hook, `Carousel` read `scrollLeft` **and** wrote a custom property in the same read phase. Both are fixed on `main`, and nothing failed, nothing warned, and no test could see it: the pixels were right, the layout work was not.

Checked against the two defects it exists for. Linting the pre-fix files of `39dad09d` reports **both** — `Carousel` directly, `Cursor` through the one-level `this.<name>(…)` resolution from its `smoothTo()` subscriber. The fixed files are silent, and stripping `@write` from the fixed `Cursor` makes it report again, so the exemption is load-bearing rather than incidental.

Two mechanisms remove the false positives, and both are general rather than special cases: the armed state clears at **any** nested function boundary — which is exactly why a returned render function and a `$write()` callback are correct by construction — and a `this.<name>(…)` call resolves **one** level into that method unless it carries `@read` or `@write`.

**Widened beyond gap 43, on evidence.** `scroll.ts` coalesces into `defaultScheduler.read()` (`scroll.ts:133-139`) and `scroll-progress.ts` emits from inside a read task too, so `scrolled()` and `scrolledInView()` carry the identical hazard and are armed alongside `ticked()`. The other six services — resize, pointer, key, mutation, in-view, drag — emit straight from their listener or observer, in no scheduler phase at all, so their hooks are **not** read-phase contexts and are deliberately absent.

## What the rules found in real code

**`prefer-instance-scheduler` — four hits, one genuine defect, fixed here.** The ScrollAnimation family reaches past the instance lane at four call sites, and both sites carry a comment justifying it. Both comments are wrong. They claim `$destroy()` cancels the instance's pending tasks *after* the cleanups, so a render scheduled from teardown would be dropped. It cancels them *before*: `$destroy()` takes the pending set, cancels it, installs a fresh one, and only then runs the cleanups and `destroyed()` — `Base.ts:1028` says so in as many words, because tasks scheduled by cleanup are meant to remain active. So the instance lane was available all along, and using the global one meant a task nothing could ever cancel: the leak class `$read()`/`$write()` exist to prevent. Fixed in `fix(v4): schedule the boundary render on the instance lane`, with the stale comments corrected. 1382 tests still pass, unchanged.

**`no-dollar-prefix` — two hits, both false positives, so the rule was dropped from the v4 set.** It was my own speculative inclusion, not part of the ask. Its two hits are `registry.spec.ts:176`, a test double *overriding* the framework's own `$mount()` to make it throw, and `DataBind.ts:136`'s `get $data()`, whose name is deliberate public API — it is the identifier userland expressions are handed, as in `data-option-compute="$data.first"`. One speculative benefit against two real false positives is not a trade worth making. No inline disables were added anywhere.

**Everything else is silent** over the whole of `packages/v4`, including `migration/`. `no-write-in-read-phase`, `no-options-assignment`, `option-default-factory`, `no-conflicting-negated-option` and the v4 deprecation set report nothing on current code.

## `require-super-mounted`: evaluated and refused

PR #850 moved core's service mixins off `mounted()` onto `$mount()`/`$destroy()` precisely to remove the trap, and its commit body says so: *"The ask on the roadmap was a diagnostic. This removes the trap instead."* For a core mixin there is now nothing to chain, so a rule firing on every `mounted()` in a `with*()` subclass would be pure noise — and worse, it would push authors to write a `super.mounted()` call that does nothing, which is a false signal of safety.

The trap survives only for a *userland* mixin which puts its work in `mounted()`. A single-file, type-free linter cannot tell that case from the core one: the mixin's body is in another file, and `withFoo` and `withRaf` are indistinguishable at the call site. A rule that cannot separate the case it must catch from the case it must ignore is a rule that only produces false positives. Refused.

## Not built, deliberately

- **`@read`/`@write` overridden by a subclass.** The base and the subclass live in different files, so a single-file linter cannot see the pair. Agreed with the brief after checking.
- **Any `data-ref` spelling rule.** The two halves live in JS and in Twig. Agreed.

## Two things that contradicted the brief

- **`this.$root` is not a v4 anchor** — `$root` does not exist in v4 at all (`DESIGN.md:33`), so it is absent from `no-write-in-read-phase`'s anchor set, which is `this.$el`, `this.$refs.*`, `document`, `window` and locals initialised from those. It stays in the *v3* deprecation list where it belongs.
- **`$options` is not a getter-only view at the top level.** It is a plain writable field (`Base.ts:899`); it is each *option property* on it that is a getter with no setter (`Base.ts:693-696`). The rule therefore reports `this.$options.x = …` and `Object.assign(this.$options, …)`, which do throw, and leaves `this.$options = …` alone, which does not.

## A note on the plugin's authoring shape

`createOnce()` runs once for the whole run while `context` tracks the file being linted, so reading `context.options` in the setup body captures whichever file happened to be first and applies it to every other file. This bit the `version` option and would bite any future rule with options or per-file state. Options are now read per file at report time, and `no-write-in-read-phase` rebuilds its state on `Program`.

Also worth recording: the RuleTester's parser is espree, which reads neither decorators nor TypeScript, so no spec fixture can contain `@write` or a type annotation. The helpers reading them are unit tested against hand-built nodes in `ast.spec.ts`, and the decorator paths are covered end to end by oxlint over `packages/v4`, which parses both.

The v4 set is documented in the plugin's README rather than in `packages/docs`, which is the v3 documentation site.

## Review round

Rebased onto `origin/main` (through #854) with no conflicts. Three changes came out of it.

**`isComponentClass()` was too loose.** It accepted any class with any superclass, so `class Store extends Error` read as a component — and `prefer-instance-scheduler` would then tell it to call a `this.$read()` it does not have, which is wrong advice rather than noise. A list of known bases is not the fix, because `extends AbstractCarouselChild`, `extends Indexable` and `extends AbstractPrefetch` are the common shape in `@studiometa/ui` and a single-file linter cannot follow the chain. The test is now a superclass **plus** a second signal in the same class body: a `@component` decorator, a `static config`, or use of the framework surface off `this`. Lifecycle and service hook names are deliberately excluded from that surface — they are what `no-write-in-read-phase` keys on, so counting them would make the test circular.

**The loose version was also wrong in the other direction, which only the rebase revealed.** `Base` declares `static config = { name: 'Base' }` and defines the whole surface, so it read as a component and produced two live `prefer-instance-scheduler` warnings inside `Base.ts` — telling the class that *defines* `$read()` to call it. Requiring the superclass keeps it out, since `Base` extends nothing.

**#854 invalidated a carve-out.** `no-options-assignment` deliberately left `this.$options = …` alone because `$options` was a plain writable field. It is now defined non-writable along with `$el`, `$id` and `$refs`, so replacing the view throws too; it is reported, with its own message, because the failure is a different one. `$el`, `$id` and `$refs` are now in the same position and are **not** covered here — a rule named for options should not quietly grow to police four properties. Worth its own rule.

`props.spec.ts`, added by #854, is a run of `@ts-expect-error` lines proving the type rejects those writes. A fixture written wrong on purpose should not be reported, so the five new rules join the js-toolkit rules the spec-file override already turns off.

Both defects of `39dad09d` are still reported after all of this, so the tightening cost no coverage where it counts, and `packages/v4` is otherwise clean.

## Verification

| Command | Result |
| --- | --- |
| `npm run test:eslint-plugin` | 36 files, 55 tests passed — exit 0 |
| `npm run lint` | passed, 2 pre-existing `unicorn/no-useless-spread` warnings — exit 0 |
| `packages/v4`: `npm run lint:types` | passed — exit 0 |
| `packages/v4`: `npm run test` | 90 files, 1386 tests passed — exit 0 |


titouanmathis:
Superseded by #856, which carries the same commits from `feat/v4-lint-rules`. This one was opened from the agent worktree's own branch name, and that name would have ended up in the permanent merge commit.

